PR #1, opened 2021-05-18
styles/constants: adding constants/libs to use them for styling later:

khamui:
Added FONTS, COLORS and SPACINGS (based on values of adhocracy+)
Initial setup. To be extended modified

### DIFF
--- a/theme/colors.js
+++ b/theme/colors.js
@@ -1,0 +1,28 @@
+export const COLORS = {
+  // brand
+  primary: '#2d40cc',
+  secondary: '#90efe4',
+  tertiary: '#008dc5',
+
+  // priority
+  success: '#a3ef90',
+  info: this.primary,
+  warning: '#ffc107',
+  danger: '#a72b1e',
+
+  // utility
+  paper: {
+    main: '#fff',
+    inverted: this.text.main
+  },
+  text: {
+    main: '#1a1a1a',
+    inverted: this.paper.main
+  },
+  grey: {
+    extralight: '#d8d8d8',
+    light: '#4d4d4d',
+    medium: '#666',
+    dark: '#464a4c',
+  },
+};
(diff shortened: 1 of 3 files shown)


khamui:
not really sure what naming is consistent, descriptive and flexible at the same time. kinda cryptic now 😵‍💫

Rineee:
> not really sure what naming is consistent, descriptive and flexible at the same time. kinda cryptic now 😵‍💫

Haha, we would need @phillimorland for this :) But I think I would prefer the sm-md-lg naming, because it is consistent with our other naming.. what do you think? Or do you think its not flexible enough?

theme/colors.js
```diff
@@ -0,0 +1,28 @@
...
+    main: '#1a1a1a',
+    inverted: this.paper.main
+  },
+  grey: {
+    extralight: '#d8d8d8',
+    light: '#4d4d4d',
+    medium: '#666',
+    dark: '#464a4c',
+  },
+};
```
goapunk, line 28: not sure how this is handled in the other projects, but maybe add a newline at the end of this line (same for the other two files)?

fuzzylogic2000 (commented):
Nice! But do we need this many differnt spacings and font sizes? Can we start with a smaller set and only add more when we absolutely need to?

theme/colors.js
phillimorland: our linter automatically saves a blank last line, I can look into our linter configs and see what can be reused here.

khamui: i can add the rule `eol-last` to it, that will make sure any file ends with a newline (think that is what we used in our other linting as well)

khamui:
okay will reduce the list. makes sense to start with as least as possible!

phillimorland (approved):
Nice! :)

PR #4, opened 2021-05-21
WIP: ideas: add container and dummy idea title add text field

fuzzylogic2000:


containers/Ideas/IdeaCreate.js
```diff
@@ -0,0 +1,27 @@
...
+    <View style={styles.inputGroup}>
+      <TextInput
+        placeholder="good idea"
+        style={styles.TextInput}
+        onChangeText={textInputHandler}
+        value={enteredIdeaTitle}
+      />
+      <Button
+        title="ADD"
+        onPress={props.onAddIdea.bind(this, enteredIdeaTitle)}
```
khamui, line 22: to avoid using `this`, we could use `() => props.onAddIdea(enteredIdeaTitle)` ? would that work?

khamui (commented):
nice job!

containers/Ideas/IdeaCreate.js
fuzzylogic2000: Cool! I will try!

PR #5, opened 2021-05-25
module/ideas: preparing structure and aligning to wireframes (wip)

khamui:
Dividing elements into container and presentationals.
Prestyling (to be changed) according to structure of wireframes.
ToDo: Refactoring, Icon handling, styling to meet designs (wip).


containers/Ideas/IdeasListItem.js
```diff
@@ -0,0 +1,29 @@
+import React from 'react';
+import { View, Text, TouchableOpacity } from 'react-native';
+import { styles } from './IdeasListItem.styles';
+
+import { ButtonCounter } from '../../components/buttonCounter';
+
+export const IdeasListItem = (props) => {
+  console.log(props);
```
Rineee, line 8: Could you remove this?

Rineee (changes requested):
Really cool!
Could you maybe somehow squash your commits, at least the first and second? So the mocked data is gone :)
And there are still some files without newlines at the end, could you add them, please?

The rest looks good to me, so we can go on from there!

containers/Ideas/IdeasListItem.js
khamui: ahh, thanks..maybe i'll add another eslint rule.

khamui:
okay now should be fixed with linting rules and squashed. additionally i removed the sample data.

Rineee (approved):

PR #6, opened 2021-05-26
Idea IdeasListItem - adding icons

phillimorland:


phillimorland:
This is currently importing icons directly, I think it might be better to have this as a dependency but I don't know enough about it yet, also didn't know it could work like this so thats cool! :)

khamui:
Pretty cool @phillimorland !

khamui:
ah just saw now that you installed a new dependency `react-native-elements`. Do we need that? Or could we build our own button component which allows icons?

@phillimorland

phillimorland:
@khamui I think we totally can, but i'm not sure if it's worth it at this point, for accessibility and ease for first concept this should be ok? I will add to the pad to look into how to make custom buttons and ensure accessibility though!

khamui:
actually while integrating this into my code, i think it may be better to use this extra dependency. Exactly as you said because of accessiblity and also cross plattform mobile gesture touch behavior it is maybe better then inventing the wheel again. So lets use it

PR #9, opened 2021-05-27
Add sentry integration

goapunk:


fuzzylogic2000 (approved):
Looks good!
Is there anything (else) we need to do if we want to get a new version up to expo?

PR #10, opened 2021-05-28
Ks 2021 05 add central api

Rineee:
I already added a method for post requests and an example for adding a live question, because it doesnt need authentification, see usage in comments

### DIFF
--- a/BaseApi.js
+++ b/BaseApi.js
@@ -1,0 +1,65 @@
+const baseUrl = 'https://aplus-dev.liqd.net/api'
+
+const endpoints = {
+  ideas: baseUrl + '/modules/$moduleId/ideas',
+  liveQuestions: baseUrl + '/modules/$moduleId/interactiveevents/livequestions/'
+}
+
+const makeGetRequest = (url) => {
+    return fetch(url)
+      .then((response) => response.json())
+      .then((unpackedData) => {
+        return unpackedData
+      })
+      .catch((error) => {
+        console.error(error);
+      });
+}
+
+const makePostRequest = (url, data = {}) => {
+     return fetch(url, {
+       method: 'POST',
+       headers: {
+           'Accept': 'application/json',
+           'Content-Type': 'application/json'
+       },
+       body: JSON.stringify(data)
+       })
+       .then((response) => response.json())
+       .then((responseData) => {
+         return responseData
+       })
+       .catch((error) => {
+         console.error(error);
+       });
+}
+
+
+const API = {
+    getIdeas(moduleId) {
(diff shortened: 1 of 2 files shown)


BaseApi.js
```diff
@@ -0,0 +1,65 @@
...
+
+const endpoints = {
+  ideas: baseUrl + '/modules/$moduleId/ideas',
+  liveQuestions: baseUrl + '/modules/$moduleId/interactiveevents/livequestions/'
+}
+
+const makeGetRequest = (url) => {
+    return fetch(url)
+      .then((response) => response.json())
+      .then((unpackedData) => {
```
khamui, line 11: oh just saw i left this unnecessary line break (probably when trying something).
Here you could make a one line as the line above.

BaseApi.js
```diff
@@ -0,0 +1,65 @@
...
+     return fetch(url, {
+       method: 'POST',
+       headers: {
+           'Accept': 'application/json',
+           'Content-Type': 'application/json'
+       },
+       body: JSON.stringify(data)
+       })
+       .then((response) => response.json())
+       .then((responseData) => {
```
khamui, line 29: we could return that newly created data here.

BaseApi.js
```diff
@@ -0,0 +1,65 @@
...
+    and see that question has been posted
+
+    */
+    postLiveQuestion(moduleId, data) {
+      const url = endpoints.liveQuestions.replace(/\$(\w+?)\b/g, moduleId)
+      return makePostRequest(url, data)
+    }
+};
+
+export default API;
```
khamui, line 65: the newline at the end is missing. Linting should give a hint.

khamui (commented):
Very cool @Rineee !

BaseApi.js
Rineee: Right!

BaseApi.js
Rineee: Yeah I didnt really know what to do here, because we are not really using it yet. So you mean `return responseDate` here instead?

BaseApi.js
Rineee: Will remove it then!

BaseApi.js
khamui: yes exactly

BaseApi.js
Rineee: Hm, thats giving me an error `Unexpected token` :woman_shrugging: So I'll leave it like this!

fuzzylogic2000 (approved):
Nice!

PR #11, opened 2021-05-31
Kl 2021 05 fake idea create

fuzzylogic2000:
@phillimorland This is it...
There is some not used stuff in there a as well like the state on the idea create page. And the styles. Should I remove it or do you want to replace it by working things?


fuzzylogic2000:
@phillimorland Maybe it's best if I make another PR really just adding the navigation and an empty create page for us to merge and for you to go on? Yes, I will do that!

PR #12, opened 2021-05-31
ideas: add navigation to idea form

fuzzylogic2000:
So, I think, this is sth. we can merge and then you can add the actual form @phillimorland !

### DIFF
--- a/containers/Ideas/IdeaCreate.js
+++ b/containers/Ideas/IdeaCreate.js
@@ -1,0 +1,24 @@
+import React, { useState } from 'react';
+import { Button, TextInput, View } from 'react-native';
+import { styles } from './Idea.styles'
+
+export const IdeaCreate = props => {
+
+  const pressHandler = () => props.navigation.pop();
+
+  return (
+    <View>
+      <TextInput
+        placeholder="This is my great idea"
+        style={styles.TextInput}
+      />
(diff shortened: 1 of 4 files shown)


containers/Ideas/IdeaCreate.js
```diff
@@ -0,0 +1,24 @@
+import React, { useState } from 'react';
+import { Button, TextInput, View } from 'react-native';
+import { styles } from './Idea.styles'
+
+export const IdeaCreate = props => {
+
+  const pressHandler = () => props.navigation.pop();
```
fuzzylogic2000, line 7: This is to go back to the idea list once the idea is added. Later we need to use it at the same time with posting the idea

containers/Ideas/IdeasPage.styles.js
```diff
@@ -4,6 +4,12 @@ export const styles = StyleSheet.create({
   container: {
     flex: 1,
     backgroundColor: 'white',
-    padding: 32
+    padding: 32,
+    paddingBottom: 70
   },
+  absoluteBottomBtn: {
```
fuzzylogic2000, line 10: Yeah, and I didn't really get why the button was already where it's meant to be without any layouting (apart from padding). I also didn't get it to be full width without breaking all the styling. :(

phillimorland (approved):
Awesome! :)

PR #13, opened 2021-05-31
config/linting: Adding deps for pre-commit linting - husky & lint-staged

khamui:


khamui:
@phillimorland I checked preLint and saw that this also needs another dependency (pre-commit). So I am using Husky, which we use in aplus as well  and just configure eslint (instead of style-lint and py-lint).

phillimorland (approved):
Nice!

PR #16, opened 2021-06-02
Pm 2021 06 more form components

phillimorland:


fuzzylogic2000 (approved):
Wow! It looks so awesome! And it works!
(one very small thing for later: should the whole labels be clickable to select/deselect them?)

phillimorland:
@fuzzylogic2000 yeah the clickable label was discussed yesterday during standup and it was draining alot of time so @mcastro-lqd said it was ok to just have tick boxes for now but if we want to make our application for release by end of year then we maybe need to reconsider that but as this is a proof of concept then I would say we can worry about it in the next sprint. I'm also not totally sure on the dropdown package but the native community one was just causing issues so I'm keeping both in mind.

PR #27, opened 2021-06-07
Ks 2021 06 add login screen

Rineee:
There is now a login screen, but once you logged in with a valid account on aplus dev, it should store the token an no more login is required..

### DIFF
--- a/containers/Auth/StartUpScreen.js
+++ b/containers/Auth/StartUpScreen.js
@@ -1,0 +1,28 @@
+import React, { useEffect } from 'react';
+import { ActivityIndicator, View } from 'react-native';
+import AsyncStorage from '@react-native-async-storage/async-storage';
+
+import { styles } from '../Ideas/Idea.styles';
+import { COLORS } from '../../theme/colors';
+
+export const StartUpScreen = (props) => {
+
+  useEffect(() => {
+    const tryLogin = async () => {
+      const authToken = await AsyncStorage.getItem('authToken');
+      if (!authToken) {
+        props.navigation.navigate('Auth');
+      }
+      else {
+        props.navigation.navigate('Ideas');
+      }
+    };
+    tryLogin();
+  }, []);
+
+  return (
+    <View style={styles.container}>
+      <ActivityIndicator size='large' color={COLORS.grey.medium} />
+    </View>
+  );
+};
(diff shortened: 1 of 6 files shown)


containers/Auth/StartUpScreen.js
```diff
@@ -0,0 +1,28 @@
...
+
+export const StartUpScreen = (props) => {
+
+  useEffect(() => {
+    const tryLogin = async () => {
+      const authToken = await AsyncStorage.getItem('authToken');
+      if (!authToken) {
+        props.navigation.navigate('Auth');
+      }
+      else {
```
khamui, line 16: just a little thing, but we should align on the same brace style for if-statements.
I also like the stroustrup brace style. @Rineee @phillimorland @fuzzylogic2000 if you don't mind using this style i will add a check to eslint?

khamui (commented):
Very cool! Logging in works ⚡️
just some small things, i commented below.

khamui:
Actually my comments are not relevant to the functionality. It works, so if you want to postpone these little changes to the next sprint let me know so i merge immediately @Rineee .

containers/Auth/StartUpScreen.js
fuzzylogic2000: So this is the stroustrup brace style? Everything sound complicated for js. :rofl: Yes, I like it (looks a bit like python :stuck_out_tongue:), let's use it!

navigation/IdeaNavigator.js
```diff
@@ -13,4 +15,18 @@ const IdeaNavigator = createStackNavigator({
...
+const AuthNavigator = createStackNavigator({
+  Authentication: LoginScreen
+});
+
+const StartUpNavigator = createStackNavigator({
+  StartUp: StartUpScreen
+});
+
+const MainNavigator = createSwitchNavigator({
+  StartUp: StartUpNavigator,
```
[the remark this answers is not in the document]
Rineee: Yes, you are right! It somehow didnt work when I first tried it, thats why I added those StackNavigators around, but I just tried again and it works without, dont know what I was doing :see_no_evil:

fuzzylogic2000 (approved):
Whoop whoop!

containers/Auth/StartUpScreen.js
khamui: haha 😝

PR #29, opened 2021-06-07
module/ideas: using module api and phase information

khamui:


fuzzylogic2000 (changes requested):
Very nice! 
Only thing: it can't handle when there is no active phase. Maybe for now we don't show any phase then? (and add a task/issue to ask about that). 
Only showing the idea add button when the collect phase is active (there's a prop in the module api for that) is still missing, but we can do that later.

containers/Ideas/IdeaProject.js
```diff
@@ -84,11 +105,13 @@ export const IdeaProject = (props) => {
                 <View style={styles.listContainer}>
                   <IdeasList ideas={ideas} {...props} />
                 </View>
-                <Button
-                  buttonStyle={styles.submitButton}
-                  title='Submit Idea'
-                  onPress={pressHandler}
-                />
+                {activePhase && (
```
fuzzylogic2000, line 108: It should only appear in idea collect, not in rating phase. I meant this field for it: https://github.com/liqd/adhocracy-plus/blob/240273e6d6f503786df40a4dab7c47bc2fbf14aa/apps/projects/serializers.py#L99
But that's for after review then!

fuzzylogic2000 (approved):
Looking good! I'll quickly try it out and merge if all fine!

PR #31, opened 2021-06-08
adding image picker from expo - initial step - limited styling, refactor needed

phillimorland:
Still to do: 

- [ ] funtion on addImageButton to make other buttons appear when pressed
- [ ] refactor code, including replace deprecated expo permissions package, update naming 
- [ ] reafactor and complete styling, including using variables


phillimorland:
so this is a bit messy and have added the list of things still to do. As the Iphone emulator has no camera it would be good to know if this works.

fuzzylogic2000 (approved):
Whoop whoop it works!

PR #35, opened 2021-06-08
module/ideas: projects without images have primary color as bg

khamui:
After checking with @mcastro-lqd we use the brand primary color (as is on the web version) as default to show in case the project does not have an image.


phillimorland (approved):
cool!

PR #36, opened 2021-06-08
Pm 2021 06 second step add image

phillimorland:
Still to do: 

- [x] funtion on addImageButton to make 'camera' and 'lirary' buttons appear when pressed
- [ ] replace deprecated expo permissions package
- [x] update naming 
- [x] reafactor and complete styling, including using variables 
- [x] cropping ratio to be checked and adjusted and image quality to ensure looks good on both mobile and desktop
- [x] move image form fields to seperate file?
- [x] need whole screen to be scrollable
- [x] description form fields char limit
- [x] stop form having capital at start
- [x] make whole checkbox clickable


phillimorland:
I think the permissions thing can be added to the technical review for next sprint as i want to see how often the other packages we are using get updated, if they are not very stable then we might need a more significant refactor then just swapping it out.

components/formFields.styles.js
```diff
@@ -1,24 +1,26 @@
...
 import { COLORS } from '../theme/colors';
+import { SPACINGS } from '../theme/spacings';
+import { SIZES, LINEHEIGHTS } from '../theme/fonts';
 
 export const styles = StyleSheet.create({
   formTitle: {
-    fontSize: 17,
-    lineHeight: 25,
+    fontSize: SIZES.md,
+    lineHeight: LINEHEIGHTS.md,
```
khamui, line 9: nice!!!

components/imageFormField.js
```diff
@@ -0,0 +1,119 @@
...
+
+export const ImagePickerFormField = () => {
+  const [image, setImage] = useState(null);
+
+  useEffect(() => {
+    (async () => {
+      if (Platform.OS !== 'web') {
+        const { status } = await ImagePicker.requestMediaLibraryPermissionsAsync();
+        if (status !== 'granted') {
+          alert('Sorry, we need camera roll permissions to make this work!');
```
khamui, line 60: what is the difference to the `Alert` above?
could we use the same?

components/imageFormField.js
```diff
@@ -0,0 +1,119 @@
...
+  };
+
+  const captureImageHandler = async () => {
+    const hasPermission = await verifyPermissions();
+    if(!hasPermission) {
+      return;
+    }
+    const image = await ImagePicker.launchCameraAsync({
+      allowsEditing: true,
+      aspect: [16, 9],
```
khamui, line 30: super detail (probably for later): we could extract this to a config variable or file and reuse it also for below (picklibraryimage).

khamui (approved):

khamui:
@phillimorland very nice state of the image adding feature!! code is very readable. I tested it and it works, except of the scrolling somehow (android). And i have some comments on tiny stuff (that definitely is for later).

But I think this state is good to merge! Let me know if you want to touch it or I can merge it.

phillimorland:
@kha i'll check the alert stuff and tag you again, rest I think we can leave for now, the scroll on android is interesting as works for me locally but I think it's becasue we have a dropdown inside a scrollable area which isn't right, I added an issue for it.

Kha:
> and tag you again

Please don't, because I'm not @khamui

fuzzylogic2000:
@khamui I will also mention you here.

fuzzylogic2000:
@khamui No, you're still not first in the list for PRs. Well, will have to comment more!

components/imageFormField.js
```diff
@@ -0,0 +1,104 @@
...
+
+export const ImagePickerFormField = () => {
+  const [image, setImage] = useState(null);
+
+  useEffect(() => {
+    (async () => {
+      if (Platform.OS !== 'web') {
+        const { status } = await ImagePicker.requestMediaLibraryPermissionsAsync();
+        if (status !== 'granted') {
+          Alert.alert('Sorry, we need camera roll permissions to make this work!');
```
khamui, line 61: nice!

PR #44, opened 2021-06-10
jd 2021 06 upgrade to react-navigation 5

goapunk:
Upgrade to the latest stable version of react-navigation and adjust to the api changes. I had to change the authentication and navigation flow because the switch-navigator has been deprecated.

### DIFF
--- a/navigation/IdeaNavigator.js
+++ b/navigation/IdeaNavigator.js
@@ -1,24 +1,37 @@
-import { createStackNavigator } from 'react-navigation-stack';
-import { createAppContainer, createSwitchNavigator } from 'react-navigation';
-
+import React from 'react';
+import { NavigationContainer } from '@react-navigation/native';
+import { createStackNavigator } from '@react-navigation/stack';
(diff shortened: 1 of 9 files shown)


navigation/IdeaNavigator.js
```diff
@@ -1,24 +1,37 @@
...
+const Stack = createStackNavigator();
 
-export default createAppContainer(MainNavigator);
+export const IdeaNavigator = (props) => {
+  const { loading, token } = useAuthorization();
+  return (
+    <NavigationContainer>
+      <Stack.Navigator
+        initialRouteName="StartUp"
+        screenOptions={{ gestureEnabled: false, headerShown: false }}
```
goapunk, line 20: this will hide the header bar with the back arrow and everything, which I think is what we want?

Rineee:
Cool! Now there is no way to go back from IdeaCreate, but I will merge anyway and add a button in my PR, will have to rebase anyway!

PR #46, opened 2021-06-11
Ks 2021 06 post idea

Rineee:
Only posts name and description of ideas.

### DIFF
--- a/containers/Auth/LoginScreen.js
+++ b/containers/Auth/LoginScreen.js
@@ -1,12 +1,13 @@
 import React, { useEffect, useState } from 'react';
-import { Alert, Button, View } from 'react-native';
+import { Alert, Button, StyleSheet, View } from 'react-native';
 import { Formik } from 'formik';
 import * as yup from 'yup';
 
 import {useAuthorization} from './AuthProvider.js';
 import API from '../../BaseApi';
-import { styles } from '../Ideas/Idea.styles';
 import { TextInputFormField } from '../../components/formFields';
+import { COLORS } from '../../theme/colors';
+import { SPACINGS } from '../../theme/spacings';
 
 export const LoginScreen = () => {
   const {signIn} = useAuthorization();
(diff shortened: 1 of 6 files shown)


fuzzylogic2000 (commented):
Reads well to me, but in the app, I still can't see any projects, so can't currently test it.

containers/Ideas/ExplorePage.js
```diff
@@ -29,7 +21,10 @@ export const ExplorePage = (props) => {
   );
 
   useEffect(() => {
-    fetchProjects();
+    const projectsListener = props.navigation.addListener('focus', () => {
```
khamui, line 24: nice! i guess this is v5.x?

khamui: ah yes the commit message says it `...make listeners work with new navigation` 💁🏾‍♂️

containers/Auth/LoginScreen.js
```diff
@@ -1,30 +1,32 @@
...
-      Alert.alert('An error occured', error[0], [{ text: 'Ok' }]);
+      Alert.alert('An error occured', error, [{ text: 'Ok' }]);
     }
   }, [error]);
 
   const handleLogin = (values) => {
     API.postLogin(values).then((response) => {
-      if (!response.token) {
-        setError(response.non_field_errors);
+      if (!(response.statusCode==200)) {
```
khamui, line 25: maybe more readable `response.statusCode !== 200`?

khamui (approved):
Really nice. Very readable indeed, i could test it and it works fine!!
@Rineee one comment, if no need to change i can merge. @fuzzylogic2000 no further objections/findings?

goapunk:
Looks goods to me. One think which is not just specific to this MR: The back button is white and therefore is not visible if the background image is white as well (e.g. in the `one more app project`).

Rineee:
> Looks goods to me. One thing which is not just specific to this MR: The back button is white and therefore is not visible if the background image is white as well (e.g. in the `one more app project`).

Yes, I also saw that. Do you want to open an issue for it?

khamui:
yeah!

PR #62, opened 2021-07-23
BaseApi: add rating functionality

Rineee:
I hope this works, couldnt really test it :grimacing: 

For both rate and changeRating the data is just the value of the rating, so data = {value: 1} or data = {value: -1}. When a rating is removed, it is handled by submitting {value: 0}.

For rating, only objectId is needed, for changeRating also the ratingId is needed, which is now serialized, see https://github.com/liqd/adhocracy-plus/pull/1373


Rineee:
@kha This is also updated now. Both rate and changeRating now also need contentTypeId as input. It will then also hopefully work for comments, so I renamed everything to more general..

khamui:
LGTM!

PR #66, opened 2021-07-28
component/menu: research and testing libraries, ending up using react…

khamui:
… elements

After reading and trying some libraries for `backdrops` and `overlays`, i found `react native elements` is what provides a similar behavior how we want it.

Nevertheless it will be not easy to get an accurate result compared to the designs. It will be a simpler version of the menu.

### DIFF
--- a/components/Menu.js
+++ b/components/Menu.js
@@ -1,0 +1,15 @@
+import React from 'react';
+import { View } from 'react-native';
+import { BottomSheet } from 'react-native-elements';
+import { MenuItem } from './MenuItem';
+
+export const Menu = (props) => {
+  return (
+    <BottomSheet isVisible>
+      <View>
+        {props.menuItems.map((menuitem, idx) => <MenuItem key={idx} item={menuitem} />)}
+        <MenuItem item={{title: 'Cancel'}} />
+      </View>
+    </BottomSheet>
+  );
+};
--- a/components/Menu.js
+++ b/components/Menu.js
@@ -1,15 +1,27 @@
 import React from 'react';
-import { View } from 'react-native';
+import { View, StyleSheet } from 'react-native';
 import { BottomSheet } from 'react-native-elements';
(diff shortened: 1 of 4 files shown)


containers/Ideas/Idea.js
```diff
@@ -15,67 +16,97 @@ export const Idea = (props) => {
...
-        <Text style={styles.text}>
-          Reference No.: {params.reference_number || 'n/a'}
-        </Text>
-      </View>
-      <View style={styles.bottomActionsContainer}>
-        <View style={styles.ratingButtons}>
-          <ButtonCounter
-            icon={<Icon name='arrow-up' size={18} />}
-            counter={params.positive_rating_count}
+    <>
```
khamui, line 47: except of last line  `<Menu...` it is just indentation

components/Menu.js
```diff
@@ -1,15 +1,27 @@
...
 };
+
+const styles = StyleSheet.create({
+  backdrop: {
+    backgroundColor: 'rgba(77, 77, 77, 0.9)'
+  },
+  list: {
+    alignItems: 'center'
+  }
+});
```
phillimorland, line 27: We are currently using seperate style sheets for components, I think we should stick with doing it one way or the other, which way do you prefer?

components/MenuItem.js
```diff
@@ -1,19 +1,67 @@
...
+  cancelContent: {
+    alignItems: 'center',
+  },
+  cancelButton: {
+    backgroundColor: 'transparent'
+  },
+  cancelButtonText: {
+    color: 'white'
+  }
+});
```
phillimorland, line 67: same as above

containers/Ideas/Idea.js
phillimorland: Doe you mean how different file is?
Also do we need the console.log?

phillimorland (commented):
just a couple of questions but other then that, looking good :)

PR #71, opened 2021-08-02
containers/ExplorePage: add a signout button to explore page story #4464

phillimorland:


phillimorland:
added on explore page for now as it's first page you go and means will not disrupt other menus until it is moved into settings

phillimorland:
We should refactor this and put it in it's own file so it can be imported but for now it can be merged

containers/Auth/AuthProvider.js
```diff
@@ -39,6 +39,11 @@ export const AuthProvider = (props) => {
       AsyncStorage.setItem('authToken', authToken);
       setState({ ...state, loading: false, token: authToken });
     },
+    signOut: async () => {
+      console.log('I was triggered during componentDidMount123');
```
goapunk, line 43: we should remove this and above as well

Rineee: Alread did that here https://github.com/liqd/zt-app/pull/78/commits/2d79284b6a28c82a8478682229367eb9080bd96a

Rineee (approved):
Yeah, really nice!

PR #73, opened 2021-08-03
components/*,container/*,theme/*style refactor: creating seperate sty…

phillimorland:
…le sheet for components and updating variables to ensure consistency

### DIFF
--- a/components/Label.styles.js
+++ b/components/Label.styles.js
@@ -1,0 +1,21 @@
+import { StyleSheet } from 'react-native';
+import { COLORS } from '../theme/colors';
+import { SIZES } from '../theme/fonts';
+import { SPACINGS, BORDERRADIUS } from '../theme/spacings';
+
+export const styles = StyleSheet.create({
+  label: {
+    alignItems: 'center',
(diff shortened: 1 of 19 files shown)


components/Label.styles.js
```diff
@@ -0,0 +1,21 @@
...
+    paddingVertical: SPACINGS.multiplyBy(.4),
+    paddingHorizontal: SPACINGS.multiplyBy(.8),
+    borderRadius: BORDERRADIUS.base,
+    marginRight: SPACINGS.multiplyBy(.5),
+    marginBottom: SPACINGS.multiplyBy(.5)
+  },
+  text: {
+    color: COLORS.text.inverted,
+    fontSize: SIZES.multiplyBy(.9375)
+  }
```
phillimorland, line 20: not sure about this font size but will leave for now

khamui:
cool! good stuff

components/Label.styles.js
khamui: yes, better than fixed pixel size i would say

PR #77, opened 2021-08-04
containers/ideas: use labels and categories from module

fuzzylogic2000:
@khamui Here's the stuff we tried!


khamui:
@fuzzylogic2000 okay i think that works. There is a warning with saying `VirtualLIsts not nested in Scrollviews`. I think it is about performance. If you want to optimize you could have a look. Think we can leave it for now as is.

Rineee (approved):
Looking very good, yeah!

PR #78, opened 2021-08-04
Ks 2021 07 send token in get

Rineee:


fuzzylogic2000 (approved):
It looks good to me! But I guess @khamui  needs to test together with his code?

khamui:
LGTM. merging and testing with my code

PR #80, opened 2021-08-05
Kt 2021 07 voting user

khamui:


fuzzylogic2000 (commented):
It looks good, but I can't vote. (I put [one more app project](https://aplus-dev.liqd.net/liqd-tests/ideas/2021-00377/) into rating phase and can vote on the web.) Can you tell me how to get it to vote?

fuzzylogic2000 (approved):
Yes! When I actually checkout your branch it works! I can even update it in the app and web and then back on the other device! So cool!

PR #81, opened 2021-08-05
components/modal: add modal, style and use as deleteModal for idea

fuzzylogic2000:
I am not sure about the reusabilty and the whole approach. But I thought any modal would have a text, an action and a cancel-thing? And if we get something else to cover, we can still change it.
Also unsure about the styling and if I would change the text-size or rather not?
I couldn't get the two lines of text centered on their own, though.

### DIFF
--- a/components/Modal.js
+++ b/components/Modal.js
@@ -1,0 +1,19 @@
+import React from 'react';
+import { View } from 'react-native';
+import { BottomSheet } from 'react-native-elements';
+import { ModalItem } from './ModalItem';
+import { styles } from './Modal.styles';
+
+export const Modal = (props) => {
+  return (
+    <BottomSheet
+      isVisible={props.isVisible}
+      containerStyle={styles.backdrop}
+    >
+      <View style={styles.list}>
+        {props.modalItems.map((modalItem, idx) =>
+          <ModalItem key={idx} item={modalItem} />)}
+      </View>
+    </BottomSheet>
+  );
+};
--- a/components/Modal.styles.js
+++ b/components/Modal.styles.js
@@ -1,0 +1,11 @@
+import { StyleSheet } from 'react-native';
+import { COLORS } from '../theme/colors';
+
+export const styles = StyleSheet.create({
+  backdrop: {
+    backgroundColor: COLORS.overlay
+  },
+  list: {
+    alignItems: 'center'
+  },
+});
--- a/components/ModalItem.js
+++ b/components/ModalItem.js
@@ -1,0 +1,38 @@
+import React from 'react';
+import { TouchableWithoutFeedback } from 'react-native';
+import { ListItem } from 'react-native-elements';
+import { styles } from './ModalItem.styles';
+
+export const ModalItem = (props) => {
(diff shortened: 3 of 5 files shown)


components/Modal.js
```diff
@@ -0,0 +1,56 @@
...
+      isVisible={props.isVisible}
+      containerStyle={styles.backdrop}
+    >
+      <View style={styles.list}>
+        <ListItem
+          Component={TouchableWithoutFeedback}
+          bottomDivider={true}
+          containerStyle={styles.modalText}
+        >
+          <ListItem.Content style={styles.modalContent}>
```
khamui, line 18: given the listitems have some things in common you could make a subcomponent something like `ModalItem`. Depending on if its not to much of conditions to set the right style classes.

containers/Ideas/Idea.js
```diff
@@ -18,7 +19,12 @@ export const Idea = (props) => {
   };
   const toggleMenu = () => setMenuVisible(!menuVisible);
   const [menuVisible, setMenuVisible] = useState(false);
-  const menuItems  = [
+  const toggleDeleteModal = () => (
```
khamui, line 22: i know i started this mess :) (changed it in a pending PR), but lets try to have an order in our react function components.

E.g. I would start with variable and state definitions, followed by function definitions. Maybe something like the following?

`const {params, createdDate} = props.route.params;`
`const {menuVisible, setMenuVisible} = useState(false);`
`const {deleteModalVisible, setDeleteModalVisible} = useState(false);`
`const toggleMenu = () => ....`
`const toggleDeleteModal = () => ....`

components/Modal.styles.js
```diff
@@ -0,0 +1,40 @@
+import { StyleSheet } from 'react-native';
```
khamui, line 1: the font sizes are looking slightly bigger than from the menu component.

khamui (commented):
Very nice!!! Working fine, just some little suggestions for improvements.

components/Modal.js
khamui: so if that is to much for now. leave it as is (because its working fine!)

fuzzylogic2000: Hm, they share nearly no styling. But might still be more readable. Will try it!

containers/Ideas/Idea.js
fuzzylogic2000: Ah, yes! I like rules I can stick to. :)

components/Modal.styles.js
fuzzylogic2000: Yeah, before they looked so small. Will remove the sizing again and see if @mcastro-lqd complains. :p

components/ModalItem.js
```diff
@@ -0,0 +1,38 @@
+import React from 'react';
```
khamui, line 1: nice!

fuzzylogic2000: Cool, so can we merge?

khamui:
Yes! looking very good.

PR #88, opened 2021-08-11
Ks 2021 08 post category and labels

Rineee:
Finally ... only works with https://github.com/liqd/adhocracy-plus/pull/1394

### DIFF
--- a/containers/Ideas/IdeaCreate.js
+++ b/containers/Ideas/IdeaCreate.js
@@ -6,7 +6,6 @@ import * as yup from 'yup';
 import AsyncStorage from '@react-native-async-storage/async-storage';
 import Icon from 'react-native-vector-icons/SimpleLineIcons';
 import IconSLI from 'react-native-vector-icons/SimpleLineIcons';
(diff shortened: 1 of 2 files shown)


containers/Ideas/IdeaCreate.js
```diff
@@ -6,6 +6,7 @@ import * as yup from 'yup';
 import AsyncStorage from '@react-native-async-storage/async-storage';
 import Icon from 'react-native-vector-icons/SimpleLineIcons';
 import IconSLI from 'react-native-vector-icons/SimpleLineIcons';
+import DropDownPicker from 'react-native-dropdown-picker';
```
fuzzylogic2000, line 9: I guess github actions is right here: should be removed if it's not used.

fuzzylogic2000 (approved):
Nice! Works well! Just the one tiny thing, we can also do later.

containers/Ideas/IdeaCreate.js
Rineee: Its removed!

fuzzylogic2000 (approved):
Nice!

PR #89, opened 2021-08-11
containers/ideas: delete idea

fuzzylogic2000:
depending on: https://github.com/liqd/adhocracy-plus/pull/1395

No error handling so far, but should be ok as button only shown when deletion allowed.

### DIFF
--- a/containers/Ideas/Idea.js
+++ b/containers/Ideas/Idea.js
@@ -1,5 +1,5 @@
 import React, { useState, useEffect } from 'react';
-import { View, Text, Image, ScrollView } from 'react-native';
+import { Alert, View, Text, Image, ScrollView } from 'react-native';
 import { Button } from 'react-native-elements';
 import { styles } from './Idea.styles';
 import Icon from 'react-native-vector-icons/SimpleLineIcons';
@@ -13,7 +13,7 @@ import { Comments } from '../Comments/Comments';
 import { TouchableOpacity } from 'react-native';
 
 export const Idea = (props) => {
-  const {params, createdDate} = props.route.params;
+  const {params, createdDate, moduleId} = props.route.params;
   const [menuVisible, setMenuVisible] = useState(false);
   const [deleteModalVisible, setDeleteModalVisible] = useState(false);
   const [voteUp, setVoteUp] = useState(params.positive_rating_count);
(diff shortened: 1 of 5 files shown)


containers/Ideas/Idea.js
```diff
@@ -34,13 +34,13 @@ export const Idea = (props) => {
...
       action: () => toggleDeleteModal(),
-      isAllowed: params.has_changing_permission
+      isAllowed: params.has_deleting_permission
     },
     {
       title: 'Report',
       icon: 'flag',
       action: () => console.log('Report'),
-      isFirst: !params.has_changing_permission,
+      isFirst: !params.has_changing_permission && !params.has_deleting_permission,
```
khamui, line 43: nice!!

khamui (commented):
pretty cool! Deleting works!

Somehow the recent deleted idea is still listed on the project page. Do we want to tackle that now? Or shall I create an issue?

fuzzylogic2000:
> Somehow the recent deleted idea is still listed on the project page. Do we want to tackle that now? Or shall I create an issue?

I needed quite a while to get that working and it was fine locally for me. And it should work, right? I don't use the ideaProject page from the navigator stack, but go to it and by that reload the page?!

khamui:
> > Somehow the recent deleted idea is still listed on the project page. Do we want to tackle that now? Or shall I create an issue?
> 
> I needed quite a while to get that working and it was fine locally for me. And it should work, right? I don't use the ideaProject page from the navigator stack, but go to it and by that reload the page?!

yes it should work by using `goBack()`, but it might be related with that listener `focus` in the project page, which somehow doesn't update the list. But if it worked for you we can merge and test on dev?

Not sure if that helps:
https://github.com/react-navigation/react-navigation.github.io/issues/191#issuecomment-641018588

containers/Ideas/Idea.js
```diff
@@ -114,6 +114,16 @@ export const Idea = (props) => {
       setVoteDown(0);
     }
   };
+
+  const deleteIdea = () => {
+    AsyncStorage.getItem('authToken')
+      .then((token) => API.deleteIdea(moduleId, params.pk, token))
+      .then(() => {
+        Alert.alert('Your idea was deleted.', 'Thank you for participating!',  [{ text: 'Ok' }]);
+        props.navigation.navigate('IdeaProject');
```
fuzzylogic2000, line 123: @khamui But I don't use goBack(), but this. goBack didn't reload the list, but for me this did. Maybe @Rineee also needs to try and decide if it works or not. :p

khamui: @fuzzylogic2000 oh I tried an old version locally. With the new version all is good and working fine!

PR #92, opened 2021-08-16
app/voting: fix voting logic & highlight user vote

khamui:
- more robust on click rage
- equivalent vote numbers of detail and list item
- bugfix: increment/decrement instead of fixed values of 0 and 1

### DIFF
--- a/theme/fonts.js
+++ b/theme/fonts.js
@@ -28,7 +28,8 @@ export const LINEHEIGHTS = {
 export const FONTWEIGHT = {
   semiBold: '500',
   bold: '600',
-  boldxl: '800'
+  boldxl: '800',
+  boldxxl: 'bold'
 };
 
 export const LETTERSPACING = {
--- a/components/ButtonCounter.js
+++ b/components/ButtonCounter.js
@@ -9,12 +9,22 @@ export const ButtonCounter = (props) => {
       onPress={props.onPress}
       disabled={props.disabled}
     >
-      <Text style={props.disabled ? styles.textDisabled : styles.text}>
+      <Text style={
+        props.disabled
+          ? styles.textDisabled
+          : props.highlight
+            ? styles.textHighlight
+            : styles.text
+      }>
         {props.counter}
       </Text>
-      <Text
-        style={props.disabled ? styles.iconDisabled : styles.icon}
-      >
+      <Text style={
+        props.disabled
+          ? styles.iconDisabled
+          : props.highlight
+            ? styles.iconHighlight
+            : styles.icon
+      }>
         {props.icon}
       </Text>
     </TouchableOpacity>
--- a/components/ButtonCounter.js
+++ b/components/ButtonCounter.js
@@ -12,18 +12,22 @@ export const ButtonCounter = (props) => {
       <Text style={
         props.disabled
           ? styles.textDisabled
-          : props.highlight
-            ? styles.textHighlight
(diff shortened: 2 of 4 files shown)


BaseApi.js
```diff
@@ -31,21 +31,17 @@ const getHeaders = (token) => {
...
-    .then(values => {
-      return {'statusCode': values[0], 'data': values[1]};
-    })
-    .catch(error => console.error(error));
+  });
+
+  const statusCode = response.status;
+  const payload = response.json();
+  const values = await Promise.all([statusCode, payload]);
+  return {'statusCode': values[0], 'data': values[1]};
```
Rineee, line 44: Is this some different functionality than before or is it just syntax? As far as I understood, the 'await' and 'then' are kind of doing the same, is that right?

khamui: Yes it is basically the same. Just syntactical sugar 😶‍🌫️
i thought i write `async/await` because the whole data passing chain i wrote it that way. Personally i think it is more readable than promise notation

Rineee: Ok, I dont mind, we can also use that notation, if you prefer it...But to be consistent, we should then also change it in all other places, so also for makeGetRequest and makePutRequest and all the methods that use it. 
And... is there a way to catch an error and log it?

khamui: hm, okay maybe then i will  change that one to promise notation. think my slightly tendency to async await does not justify the change on a lot of spots in our code..hehe. especially we would need to re-test and verify all spots are still working after change.

with async/await we would need to use `try catch`. which also makes it a bit less readable at the end.

khamui:
@Rineee okay. i changed it for the baseApi. on the consuming component i am still using async/await, changing now would be a lot and probably breaking things.

additionally i had in some rare cases negative counts. i added hotfix for it, probably not the most elegant way but it works. Hope that is fine for now.

fuzzylogic2000:
> additionally i had in some rare cases negative counts. i added hotfix for it, probably not the most elegant way but it works. Hope that is fine for now.

That is very weird! Do you mean it came from the API? Or is it happening in the app code? IMHO, we should find out where it's coming from and fix the real cause.

khamui:
> > additionally i had in some rare cases negative counts. i added hotfix for it, probably not the most elegant way but it works. Hope that is fine for now.
> 
> That is very weird! Do you mean it came from the API? Or is it happening in the app code? IMHO, we should find out where it's coming from and fix the real cause.

okay, i would need to investigate that a little more then.

khamui:
@Rineee @fuzzylogic2000 okay it was fairly simple. Reason for negative counts is if you initially doubleclick vote. that leads to a calculation of `0 - 1`, instead of `0 + 1 - 1`. A guard makes sure that that won't happen anymore.

theme/fonts.js
```diff
@@ -28,7 +28,8 @@ export const LINEHEIGHTS = {
 export const FONTWEIGHT = {
   semiBold: '500',
   bold: '600',
-  boldxl: '800'
+  boldxl: '800',
+  boldxxl: 'bold'
```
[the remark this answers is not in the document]
khamui: okay having a look tomorrow first thing

khamui:
fixing the vote colors (red for down vote missing)

fuzzylogic2000 (approved):
Hooray! Looks good and works!

PR #97, opened 2021-08-17
app/comment: excerpt only if needed and sub comments with read more

khamui:


Rineee (approved):
Yes, really cool, works perfectly now!

PR #100, opened 2021-08-19
app/voting: fix counter of intial voting

khamui:


Rineee (changes requested):
Oh no, I am super sorry, but I found another issue :see_no_evil: 

So my use case was an idea with 2 ratings, one up vote from user and one down vote from admin. Logged in as admin and removing my vote (so clicking on down) is fine, but then when readding down vote, the up vote from user vanishes (or at least the display of it). So it then shows me 0 up votes and 1 down vote. When I go back to list view though, it shows the right amount (1 up and 1 down).

khamui:
> Oh no, I am super sorry, but I found another issue 🙈
> 
> So my use case was an idea with 2 ratings, one up vote from user and one down vote from admin. Logged in as admin and removing my vote (so clicking on down) is fine, but then when readding down vote, the up vote from user vanishes (or at least the display of it). So it then shows me 0 up votes and 1 down vote. When I go back to list view though, it shows the right amount (1 up and 1 down).
> 
> But the inital display now works very well!!

no worries! i mean i could have make sure it arrives you without any issues :D Anyways, i refactored the voting logic again, as i think the approach before was too fragile (e.g. too many little bugs popped up).

Before: i kind of rewrote the logic in frontend to update the view and introduced states (upVotes/downVotes) but didn't manage to keep them up to date with the backend data (which is more reliable).

Now: After voting, i just fetch (GET) the entire idea and update the view with the numbers coming from backend. I think this makes it much more robust. I think the number of request are reasonable.

Rineee (approved):
Cool, now it works perfectly and I think this is also the way it should be done: backend provides logic and frontend displays it. Nice!

PR #102, opened 2021-08-19
app/idea-edit: fix freeze after editing idea

khamui:
@Rineee this should fix the freezing of the screen after editing an idea. Due to the menu is not closed after clicking on edit it is still there but not visible. It blocks everything. tricky. maybe the menu should close automatically to avoid this bug in future.

### DIFF
--- a/containers/Ideas/Idea.js
+++ b/containers/Ideas/Idea.js
@@ -27,7 +27,7 @@ export const Idea = (props) => {
       title: 'Edit',
       icon: 'pencil',
       action: () =>  {
-        toggleMenu();
+        setMenuVisible(false);
         props.navigation.navigate('IdeaCreate', {idea: ideaState, project: project, editing: true});
       },
       isFirst: true,
@@ -49,7 +49,7 @@ export const Idea = (props) => {
     },
     {
       title: 'Cancel',
-      action: () => toggleMenu(),
+      action: () => setMenuVisible(false),
       isCancel: true,
       isAllowed: true
     },
--- a/containers/Ideas/Idea.js
+++ b/containers/Ideas/Idea.js
@@ -26,7 +26,10 @@ export const Idea = (props) => {
     {
       title: 'Edit',
       icon: 'pencil',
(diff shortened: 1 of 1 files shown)


containers/Ideas/Idea.js
```diff
@@ -28,7 +28,10 @@ export const Idea = (props) => {
     {
       title: 'Edit',
       icon: 'pencil',
-      action: () =>  props.navigation.navigate('IdeaCreate', {idea: idea, project: project, editing: true}),
+      action: () =>  {
+        toggleMenu();
```
fuzzylogic2000, line 32: Good catch! But do we really want `toggleMenu` or do we want `setMenuVisible(false)` like when we open the deleteModal? The latter seems more robust to me.

Rineee:
> @Rineee this should fix the freezing of the screen after editing an idea. Due to the menu is not closed after clicking on edit it is still there but not visible. It blocks everything. tricky. maybe the menu should close automatically to avoid this bug in future.

Supercool! Thank you @kha! How did you find that out, that the menu is still there?

containers/Ideas/Idea.js
Rineee: I guess both work, but we should probably decide on one way. Would you then also use `setMenuVisible(false)` when clicking the Cancel button @fuzzylogic2000 ?

fuzzylogic2000: Oh right, yes, I think! I would only use the toggle when there is a button that opens and closes something. If we have sth. that only closes it, I would go for `setMenuVisible(false) `, because that's what we want there. @Rineee

khamui: yes, totally agree!

khamui:
> > @Rineee this should fix the freezing of the screen after editing an idea. Due to the menu is not closed after clicking on edit it is still there but not visible. It blocks everything. tricky. maybe the menu should close automatically to avoid this bug in future.
> 
> Supercool! Thank you @Kha! How did you find that out, that the menu is still there?

by accident ☺️

Rineee (approved):
Yeah, nice!!

PR #112, opened 2021-09-08
add Makefile with basic targets

goapunk:
Add a Makefile which makes it easier to clean the cache and run the app

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,55 @@
+.PHONY: all
+all: help
+
+.PHONY: help
+help:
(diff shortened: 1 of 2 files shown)


goapunk:
Not sure if we really need this but since there's been a lot of problems with caching and all I thought I'll give it a try

goapunk:
@phillimorland added a `make start-web` which auto-starts it with the correct chrome parameters.

phillimorland (approved):
Very nice! :)

PR #114, opened 2021-09-09
jd 2021 09 js testing

goapunk:
Add js testing framework and an initial test

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,14 +7,36 @@
     "web": "expo start --web",
     "eject": "expo eject",
     "prepare": "husky install",
-    "lintfix": "eslint --fix --ext .js,.jsx ."
+    "lintfix": "eslint --fix --ext .js,.jsx .",
+    "test": "jest",
+    "testDebug": "jest -o --coverage=false",
+    "testNoCov": "jest --coverage=false",
+    "updateSnapshots": "jest -u --coverage=false"
+  },
+  "jest": {
+    "preset": "jest-expo",
+    "collectCoverage": true,
+    "collectCoverageFrom": [
+      "**/*.{js,jsx}",
+      "!**/coverage/**",
+      "!**/node_modules/**",
+      "!**/babel.config.js",
+      "!**/jest.setup.js",
+      "!**/chrome/**"
+    ],
+    "transformIgnorePatterns": [
+      "node_modules/(?!(jest-)?react-native|react-clone-referenced-element|@react-native-community|expo(nent)?|@expo(nent)?/.*|react-navigation|@react-navigation/.*|@unimodules/.*|unimodules|sentry-expo|native-base|@sentry/.*)"
+    ]
   },
   "dependencies": {
     "@expo-google-fonts/source-sans-pro": "^0.2.0",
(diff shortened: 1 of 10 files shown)


package.json
```diff
@@ -7,13 +7,36 @@
     "web": "expo start --web",
     "eject": "expo eject",
     "prepare": "husky install",
-    "lintfix": "eslint --fix --ext .js,.jsx ."
+    "lintfix": "eslint --fix --ext .js,.jsx .",
+    "test": "jest",
+    "testDebug": "jest -o --coverage=false",
+    "testNoCov": "jest --coverage=false",
+    "updateSnapshots": "jest -u --coverage=false",
+    "type": "module"
```
goapunk, line 15: not sure if we actually need this

package.json
```diff
@@ -23,16 +46,18 @@
     "expo-status-bar": "1.0.4",
     "expo-updates": "0.8.5",
     "formik": "2.2.9",
+    "jest": "^27.1.1",
     "react": "17.0.2",
     "react-dom": "17.0.2",
-    "react-native": "https://github.com/expo/react-native/archive/sdk-41.0.0.tar.gz",
+    "react-native": "https://github.com/expo/react-native/archive/sdk-42.0.0.tar.gz",
```
goapunk, line 52: this slipped in somehow, should be fine though?

phillimorland (approved):
Mmmm so much red! ;) so it's running nicely, just need the conflicts resolved. Also maybe having make commands for running tests and making new snap shots would be cool?
Yes JS tests :D

package.json
phillimorland: It seeme to be

package.json
phillimorland: why?

goapunk: the tests wouldn't run initially and I thought this might help, I just removed it again

goapunk:
> Mmmm so much red! ;) so it's running nicely, just need the conflicts resolved. Also maybe having make commands for running tests and making new snap shots would be cool?
 I thought about it but then wasn't sure if we really need make commands since there's the npm run test|testNoCov|testDebug commands already, make won't really make it much shorter, but maybe more consistent to have them there as well?

goapunk:
resolved the conflict

phillimorland:
> > Mmmm so much red! ;) so it's running nicely, just need the conflicts resolved. Also maybe having make commands for running tests and making new snap shots would be cool?
> 
> I thought about it but then wasn't sure if we really need make commands since there's the npm run test|testNoCov|testDebug commands already, make won't really make it much shorter, but maybe more consistent to have them there as well?

The consistency is a good reason but it's also a good way to document, i'm not sure why `"jest -o --coverage=false"` if different to `"jest --coverage=false"` and we can't comment in package.json

goapunk:
pushed some changes to actually use testing-library and added test targets to the Makefile

khamui:
Pretty cool!! I want to write tests yeah!
The few test-tests are passing in local. Is it red because the coverage is too low, is it that?

Can we have jest in other repos?
If yes, can we have a little meeting about it (all that tdd is new to me)?

PR #117, opened 2021-09-09
Pm 2021 09 add category to list

phillimorland:


theme/spacings.js
```diff
@@ -11,6 +11,7 @@ export const BORDERWIDTH = {
 
 export const BORDERRADIUS = {
   none: 0,
-  base: 15,
+  base: 6,
+  md: 12,
   container: 24
```
phillimorland, line 16: adjust to be inline with zeplin

phillimorland:
@goapunk @fuzzylogic2000 should we have start adding tests everytime we work on a view do you think?

fuzzylogic2000 (approved):
Looks good!

fuzzylogic2000:
> @goapunk @fuzzylogic2000 should we have start adding tests everytime we work on a view do you think?

I would say yes! <3 tests! But let's do that starting from the next PR. We need to merge @goapunk s PR first anyway. :)

PR #131, opened 2021-09-13
Using general Source Sans Prop font family

khamui:


phillimorland (changes requested):
Looking good, there's an extra import that actions is complaining about but other then that, good to merge

components/TextSourceSans.js
```diff
@@ -0,0 +1,20 @@
+import React from 'react';
+import { Text, StyleSheet } from 'react-native';
+import { FONTS, SIZES } from '../theme/fonts';
+
```
phillimorland, line 4: actions complaint about SIZES

khamui: ah thanks for the hint!

Out of curiosity, where do you see gh actions @phillimorland?

phillimorland: I was wondering if you didn't see that, so when I click on 'add your review' and check through the code in that view, actions has comments where it's sad about things, would be good if this was included in out pre commit lint though, i'll check the config

phillimorland (approved):

components/TextSourceSans.js
khamui: ah yes, true..there are comments. and also true that this should be caught by our linting.

PR #134, opened 2021-09-13
WIP: containers/ideas: post idea image to API with formData

fuzzylogic2000:
This is posting the idea image, but gives different errors. I will try and make this nicer.
This is the main reference, I've used: https://stackoverflow.com/questions/62827501/does-formik-use-formdata-under-the-hood

But there are more things missing (please see the linked issue #141 to see related problems):

- [x] actually getting the image uri, type and name from the component and using that
- [ ] handling idea edit (showing an existing image and only put on change

--------------- Update 21. September 2021 -----------------------
We decided to keep this in it's own branch as merging would break the idea create, which is currently working. The problem seems to be posting formData instead of json. We couldn't find a solution, though.

Currently in this branch when posting an idea (with or without image), we always get this error:
`fetch error:  Network request failed`

Locally and with an hard-coded image the image post worked, though. Meaning, that the posted image was saved with the idea, even though we still got `fetch error:  Network request failed`

### DIFF
--- a/containers/Ideas/IdeaCreate.js
+++ b/containers/Ideas/IdeaCreate.js
@@ -48,6 +48,7 @@ export const IdeaCreate = props => {
   const [labels, setLabels] = useState([]);
   const [labelsChecked, setLabelsChecked] = useState([]);
   const [imageChecked, setImageChecked] = useState(false);
+  const [image, setImage] = useState();
 
   useEffect(() => {
     if (error) {
@@ -107,13 +108,33 @@ export const IdeaCreate = props => {
 
   const handleSubmit = (values) => {
     values.labels = getSelectedLabels();
+    let formData = new FormData();
+    /* FormData requires name: id */
+    // formData.append("website", "question");
+
+    /* append input field values to formData */
+    for (let value in values) {
+      formData.append(value, values[value]);
+    }
+    formData.append('image', {
+      uri: image.uri,
+      name: image.name,
+      type: image.mimeType
+    });
+
+    /* Can't console.log(formData), must
+       use formData.entries() - example:
+    for (let property of formData.entries()) {
+      console.log(property[0], property[1]);
+    } */
+
     AsyncStorage.getItem('authToken')
       .then((token) => {
         if (editing) {
           return API.editIdea(moduleId, idea.pk, values, token);
         }
         else {
-          return API.postIdea(moduleId, values, token);
+          return API.postIdea(moduleId, formData, token);
         }
       })
       //error handling is provisional and should probably go somewhere else eventually
@@ -144,6 +165,8 @@ export const IdeaCreate = props => {
       });
   };
 
+  const onSetImage = (image) => setImage(image);
+
   return (
     <VirtualScrollView
       style={styles.container}
@@ -166,11 +189,12 @@ export const IdeaCreate = props => {
           ...(editing
             ? {
               name: idea.name,
-              description: idea.description ,
+              description: idea.description,
             }
             : {
               name: '',
-              description: '' ,
+              description: '',
+
             }),
 
           labels: [],
@@ -263,7 +287,9 @@ export const IdeaCreate = props => {
               </TextSourceSans>
               {clicked &&
                 <>
-                  <ImagePickerFormField />
+                  <ImagePickerFormField
+                    onSetImage={(img) => onSetImage(img)}
+                  />
                   <CheckBoxFormField
                     field='Image Copyright'
                     name='imageCopyrightChecked'
@@ -271,6 +297,7 @@ export const IdeaCreate = props => {
                     checked={imageChecked}
                     title='I hereby confirm that the copyrights for this photo are with me or that I have received rights of use from the author. I also confirm that the privacy rights of depicted third persons are not violated.'
                   />
+
                 </>
               }
             </ImageChoiceFormFieldContainer>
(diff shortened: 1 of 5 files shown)


fuzzylogic2000:
@khamui So, this is it. :)
You'l need to change the hard-coded uri to what you get locally and in your local a+, outcomment lines 7-47 of `venv/lib/python3.6/site-packages/adhocracy4/images/validators.py`

khamui:
@fuzzylogic2000 
The error i couldn't investigate, yet! I will have a look after our afternoon meetings if I find anything

containers/Ideas/IdeaCreate.js
```diff
@@ -104,13 +105,29 @@ export const IdeaCreate = props => {
...
     values.labels = getSelectedLabels();
+    let formData = new FormData();
+    /* FormData requires name: id */
+    // formData.append("website", "question");
+
+    /* append input field values to formData */
+    for (let value in values) {
+      formData.append(value, values[value]);
+    }
+    formData.append('image', { uri: image.uri, name: image.name, type: 'image/jpeg' });
```
fuzzylogic2000, line 116: I think the type also needs to become dynamic. We might need to use a library for that, though.

fuzzylogic2000 (commented):
Nice! 
You can squash the commits, though. Now or when really done, noone needs to see the inbetween states. :)

containers/Ideas/IdeasList.js
```diff
@@ -5,7 +5,7 @@ import { IdeasListItem } from './IdeasListItem';
 
 export const IdeasList = (props) => {
   //do not pass ideas to child components
-  const {ideas, ...otherProps} = props;
```
Rineee, line 8: Haha, this was actually on purpose to remove the ideas from props and then only pass otherProps to child components :) There probably is a better way to do it? But if we remove it, we should then also remove the comment as this is wrong now.

khamui:
> Nice!
> You can squash the commits, though. Now or when really done, noone needs to see the inbetween states. :)

okay. seems to be a real bug here with the `network request failed`. I have no other idea how to get to the Idea Page after creation though. Leaving that error for now and trying to make the image types dynamic.

if you, @fuzzylogic2000, also don't have further ideas, i think we can put this state into hibernate? or what do you think?

fuzzylogic2000:
> okay. seems to be a real bug here with the `network request failed`. I have no other idea how to get to the Idea Page after creation though. Leaving that error for now and trying to make the image types dynamic.
> 
> if you, @fuzzylogic2000, also don't have further ideas, i think we can put this state into hibernate? or what do you think?

Hmm, but to put it into hibernate, we should get it into a mergeable state and do that, I think? Or do you think it is fine like that?
Maybe it also helps, when we can try on dev. To get there, you only need to merge https://github.com/liqd/adhocracy4/pull/839 and update a4 in a+. At least I hope, that's all it takes. :)

khamui:
@fuzzylogic2000 I checked again and figured out the bad request (400) happens if i crop the image to a too small value.
So it makes sense that it doesn't pass the validator with its min resolution size.

Anyway I think we can merge this state into the app. I will create an issue and document what is missing/not working with the image upload.

containers/Ideas/IdeaCreate.js
```diff
@@ -164,16 +187,17 @@ export const IdeaCreate = props => {
...
               name: '',
-              description: '' ,
+              description: '',
+
             }),
 
           labels: [],
           category: selectedCategory,
-          imageCopyrightChecked: imageChecked
+          imageCopyrightChecked: true
```
fuzzylogic2000, line 200: Oh, I only changed this to test the post. We should change it back before we merge.

fuzzylogic2000 (commented):
Cool! Works! Just one remark. :)

khamui:
Back at this again. No solution insight. It seems to be a real Bug, which a lot of people are facing using react-native and expo.

Last thing tried was the following, without success (just for me to remember what i tried):
Bildschirmfoto 2022-01-04 um 17 48 11

fuzzylogic2000:
@khamui Just an attempt of a summary so that I am fully back on board:
We are trying to send post our data as form-data instead of json, because we need that to be able to post images.
But posting the idea data with that always results in the response `fetch error: Network request failed`, which does not always mean that the post really didn't work. There is just no way so far to figure out if it worked and if not, why that was.

And you say, that problem is restricted to android and expo? It is working on iphones?
Can we develop and make it work on iphone and then figure out the android problem?

khamui:
> @khamui Just an attempt of a summary so that I am fully back on board: We are trying to send post our data as form-data instead of json, because we need that to be able to post images. But posting the idea data with that always results in the response `fetch error: Network request failed`, which does not always mean that the post really didn't work. There is just no way so far to figure out if it worked and if not, why that was.
> 
> And you say, that problem is restricted to android and expo? It is working on iphones? Can we develop and make it work on iphone and then figure out the android problem?

yes you summed it up correctly. Just some additions to that:

- the POST (201) happens always, after reload it is there
- ios/android: i tried to make it work on iphone, but didn't manage due to other issues (which i did not investigate further) -> i will tryout again today on ios

khamui:
The following list should list what to expect to be working (and what does not work) after merging this PR. I would like to create single issues for every aspect that does not work:

Android
- [x] Create idea without image
- [x] Create idea with image
- [x] Edit idea without image
- [ ] Edit image of idea with image
- [x] Edit text-only parts of idea with image

iOS
- [ ] Create idea button missing (see issue tb linked)
        - [ ] Create idea without image
        - [ ] Create idea with image
- [ ] Edit idea without image
- [ ] Edit idea with image
- [x] Edit text-only parts of idea with image

khamui:
closing manually, see cleaned up PR!

PR #135, opened 2021-09-13
app/idealist: aligning date, not depending on author name length

khamui:


khamui:
tried to make it nicer, but think for now this solution is okay. It is a compromise between phone devices and tablet devices:

Phone (short author)
phone_long-author

Phone (long author)
phone_short-author

Tablet (long author)
tablet_long-author

fuzzylogic2000 (approved):
Cool!

PR #145, opened 2021-09-21
containers/ideas: post idea image to API with formData

khamui:
app/post-image: using dynamic uri and setting image filename
app/post-image: using dynamic mime type
app/image-post: remove passing unused ideas to ideaslistitem

app/ideas: fixing bugs with props.ideas & removing babel properties


fuzzylogic2000 (approved):
Makes it work!

PR #150, opened 2021-09-22
readmore: fix show more comments and readmore

khamui:


phillimorland (approved):
Nice!

phillimorland:
@goapunk @khamui I updated the snapshot as it was failing, that ok to do right?

PR #159, opened 2021-09-28
docs: add basic debugging docs

Rineee:


khamui:
pretty cool! didn't know about adb, will try next time!

btw, think your headings are not working because of a missing space.

PR #187, opened 2021-12-27
jest: using jest 26.5, known bug for rnative/expo with v27+

### LINKED ISSUE #185: building fails: npm test fails
**URL:** -
**user:** 
**expected behaviour:** building succeeds
**behaviour:** fails because `npm test` fails
**important screensize:** -
**device & browser:** -
**Comment/Question:** It is a dev related issue, just to log here to not forget. came up while working on technical review

Screenshot?

khamui:
downgrading to jest v26... because with 27 there is a know bug that breaks npm test with react native/ expo:
closes #185 
https://github.com/expo/expo/issues/14823

and using a different strategy to run tests on ios and android: @goapunk @phillimorland not sure if this is something we want. what do you guys think?


goapunk:
Thanks for investigating this! If I read the linked issue correctly, instead of downgrading jest, upgrading to the latest expo should fix this, i.e. if we merge #188 For the iOS and Android snapshots, is there any advantage of doing so?

khamui:
> Thanks for investigating this! If I read the linked issue correctly, instead of downgrading jest, upgrading to the latest expo should fix this, i.e. if we merge #188 For the iOS and Android snapshots, is there any advantage of doing so?

okay I will try the latest expo and see if it fixes it. I faced an error when trying the minor previous version 44.01 yesterday.

About the platform specific tests (android and ios). Tbh i didn't dive very deep into what it is testing. But checking the snapshots it seems pretty much the same. So i guess for snapshot tests, indeed it doesn't really have an advantage. But the opposite, it just takes longer 👎 . I was messing around to make it the testing work and found this "feature". Lets keep in mind and see if it makes more sense to use it if we can set it per test (and not global to all tests).

https://blog.expo.dev/testing-universal-react-native-apps-with-jest-and-expo-113b4bf9cc44

khamui:
### jest version (to be used with jest-expo)
About the jest version, i did use the following command from the docs to get the version `jest-expo` is build against:
`To install the compatible version of jest-expo for your project, run: expo install jest-expo jest`

source: https://docs.expo.dev/guides/testing-with-jest/

### jest-expo config
Somehow i didn't manage to make it run with the `jest-expo` config that we used to have before (in package.json)
```
"jest": {
  "preset": "jest-expo"
}
```
But with the platform specific ones it works:
```
"jest": {
    "projects": [
        { "preset": "jest-expo/ios" },
        { "preset": "jest-expo/android"}
    ]
}
```
source: https://www.npmjs.com/package/jest-expo

Not sure if we can go with this. Think the jest version is fine, just the jest-expo config is not ideal. Do you have any suggestions on this @goapunk ?

goapunk:
@khamui adding the new config is fine, thanks!

phillimorland (approved):
looking good

PR #192, opened 2021-12-28
fix bug with dependency react-native-reanimated

khamui:
@phillimorland @goapunk I encountered another bug, while using the current versions of our dependencies.
this time it was a known bug with the package `react-native-reanimated`.

There is the option to downgrade that particular package/dependency, but people reported version issues with other dependencies. So I tried another option, which is to patch a file of that dependency (solution provided 5 days ago, the time i am writing this). And it worked.

source: https://github.com/gorhom/react-native-bottom-sheet/issues/771

For this to work I had to install another dependency called `patch-package` (gosh!). But at the end this makes it work.

`...those of us living on the bleeding edge.` - npm readme of patch-package

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,7 +11,8 @@
     "test": "jest",
     "testDebug": "jest -o --coverage=false",
     "testNoCov": "jest --coverage=false",
-    "updateSnapshots": "jest -u --coverage=false"
+    "updateSnapshots": "jest -u --coverage=false",
+    "postinstall": "patch-package"
   },
   "jest": {
     "preset": "jest-expo",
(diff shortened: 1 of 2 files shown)


goapunk:
wow, dependency hell. Nice hack - not sure what's the best option here: maintain a hack or wait for the next release (if downgrading is an option, if it makes other problems I guess we have no choice)

phillimorland:
As discussed this morning, @khamui will add either a task in taiga or github issue as reminder for us to check the state of the package during sprint, if problem has not been fixed we will need to re-assess our dependency tree.

phillimorland (approved):
installing fine :)

PR #199, opened 2021-12-29
ideas/list: if wrong backend or no ideas fetched

khamui:
If we do not have the right Backend set, it will fail! This shows a message that helps maybe to set it.

### DIFF
--- a/containers/Ideas/IdeasList.js
+++ b/containers/Ideas/IdeasList.js
@@ -1,22 +1,27 @@
 import React from 'react';
-import { View } from 'react-native';
+import {View} from 'react-native';
 
-import { IdeasListItem } from './IdeasListItem';
+import {IdeasListItem} from './IdeasListItem';
+import {TextSourceSans} from '../../components/TextSourceSans';
 
-export const IdeasList = (props) => {
+export const IdeasList = props => {
   //do not pass ideas to child components
   const {ideas, ...otherProps} = props;
+  const hasIdeas = !!ideas && ideas.length > 0;
   return (
     <View>
-      {props.ideas.map((idea) => {
-        return (
-          <IdeasListItem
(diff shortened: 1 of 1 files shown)


containers/Ideas/IdeasList.js
```diff
@@ -1,22 +1,27 @@
...
+        props.ideas.map(idea => {
+          return (
+            <IdeasListItem key={`pk${idea.pk}`} idea={idea} {...otherProps} />
+          );
+        })
+      ) : (
+        <TextSourceSans>
+          Could not fetch ideas. Do you use the correct Backend and is your
+          local adhocracy+ server running?
+        </TextSourceSans>
```
phillimorland, line 23: should we also add text for if there are just no ideas like we do on a+ with 'Nothing to show'?

fuzzylogic2000:
What is the status here?

containers/Ideas/IdeasList.js
```diff
@@ -1,22 +1,27 @@
...
 
-import { IdeasListItem } from './IdeasListItem';
+import {IdeasListItem} from './IdeasListItem';
+import {TextSourceSans} from '../../components/TextSourceSans';
 
-export const IdeasList = (props) => {
+export const IdeasList = props => {
   //do not pass ideas to child components
   const {ideas, ...otherProps} = props;
+  const hasIdeas = !!ideas && ideas.length > 0;
```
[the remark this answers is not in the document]
khamui: i don't remember why i did it that way, changed it..

containers/Ideas/IdeasList.js
khamui: hehe, yes i did it more user oriented now.

phillimorland (approved):
perfect! :)

PR #204, opened 2022-01-03
ideas/create-edit: remove console error with virtualizedlist and make…

### LINKED ISSUE #189: error: VirtualizedLists should never be nested in plain ScrollViews
**URL:** 
**user:** 
**expected behaviour:** Clicking on category dropdown (edit idea) should open without error
**behaviour:** Clicking on category dropdown shows error (see screenshot)
**important screensize:**
**device & browser:** 
**Comment/Question:** 

Screenshot?
Bildschirmfoto 2021-12-28 um 12 08 26

khamui:
… scrolling working

closes #189

### DIFF
--- a/components/VirtualScrollView.js
+++ b/components/VirtualScrollView.js
@@ -1,0 +1,18 @@
+import React from 'react';
+import {FlatList} from 'react-native';
+
+export const VirtualScrollView = props => {
+  return (
+    <FlatList
+      style={props.style}
+      contentContainerStyle={props.contentContainerStyle}
+      data={[]}
+      ListEmptyComponent={null}
+      keyExtractor={() => 'dummy'}
+      renderItem={null}
+      ListHeaderComponent={() => (
+        <>{props.children}</>
+      )}
+    />
+  );
+};
--- a/components/formFields.js
+++ b/components/formFields.js
@@ -5,6 +5,10 @@ import DropDownPicker from 'react-native-dropdown-picker';
 import { styles } from './formFields.styles';
 import { TextSourceSans } from './TextSourceSans';
 
+// Setting list mode to scrollview globally
+// FIXME: to be checked if we want this
+DropDownPicker.setListMode('SCROLLVIEW');
+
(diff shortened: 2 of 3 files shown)


components/formFields.js
```diff
@@ -5,6 +5,10 @@ import DropDownPicker from 'react-native-dropdown-picker';
 import { styles } from './formFields.styles';
 import { TextSourceSans } from './TextSourceSans';
 
+// Setting list mode to scrollview globally
+// FIXME: to be checked if we want this
+DropDownPicker.setListMode('SCROLLVIEW');
```
phillimorland, line 10: why do we want to check this?

components/VirtualScrollView.js
```diff
@@ -0,0 +1,18 @@
...
+import {FlatList} from 'react-native';
+
+export const VirtualScrollView = props => {
+  return (
+    <FlatList
+      style={props.style}
+      contentContainerStyle={props.contentContainerStyle}
+      data={[]}
+      ListEmptyComponent={null}
+      keyExtractor={() => 'dummy'}
```
phillimorland, line 11: @khamui what does the dummy thing do?

phillimorland (approved):
very nice! :)

components/VirtualScrollView.js
khamui: oh just saw this now @phillimorland , because we don't have any data this keyExtractor does nothing here.

components/formFields.js
khamui: Because its set globally, meaning all DropDownPicker will be affected. Maybe what we want anyway?

components/VirtualScrollView.js
phillimorland: danke :)

components/formFields.js
phillimorland: I guess we'll know more when we know what else we will build for the app, I think for now it's fine and we can keep the fixme in there to not forget about it

PR #216, opened 2022-01-14
containers/ideas: post idea image to API with formData

Rineee:
This is a cleaned up version of https://github.com/liqd/zt-app/pull/134/.

It fixes the image posting error thanks to @goapunk who found the problem :tada: It was, that the format of the form data was wrong (the labels need to be added each separately instead of all in one array) and then react detected the the wrong format and threw the error, but nevertheless sent the request and django did not complain about the format, but just parsed everything and sent back a response that was then ignored by react. At least thats what I understood :) It is the second commit that fixes that.

So missing is still:

- [ ] show image in editing form
- [ ] when selecting an image, previous entered data gets lost (title, idea)
- [ ] when selecting labels, previous entered data gets lost (title, idea)

I will add tasks for that!

### DIFF
--- a/BaseApi.js
+++ b/BaseApi.js
@@ -21,7 +21,6 @@ const getHeaders = (token, isFormData = false) => {
   if (token) {
     headers['Authorization'] = 'Token '+token;
   }
-  console.log(headers);
   return headers;
 };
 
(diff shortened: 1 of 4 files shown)


khamui:
Working perfectly 🙏

PR #221, opened 2022-01-17
containers/*: consolidate how we add icons to the code

phillimorland:


containers/Ideas/ExploreListItem.js
```diff
@@ -29,10 +31,7 @@ export const ExploreListItem = (props) => {
         </View>
         <View style={styles.progressContainer}>
           <TextSourceSans style={styles.progressText}>
-            <IconSLI
-              name='clock'
-              color={COLORS.grey.medium}
-            /> nur noch 12 Wochen
+            {clockIcon} time remaining {props.item.module_running_time_left}
```
phillimorland, line 34: I got this from a+ but I think I need to add it to the app serializer there right? it's defined in a4 so assume I should leave logic there?

fuzzylogic2000 (approved):
Yeah!

PR #223, opened 2022-01-17
BaseApi//Comments: add CommentForm component and post comments

Rineee:
~~Only works with https://github.com/liqd/adhocracy-plus/pull/1676/ or once that is merged.~~ Aplus is already updated.

Styling of the comment form is still missing as well as making it sticky.

To do (probably in new PR):
- [ ] change notation of if statement in Comment.useEffect
- [ ] think of better name for objectToComment
- [ ] check/fix behaviour of comment form
- [ ] only show comment form if commenting is permitted
- [ ] styling?

### DIFF
--- a/containers/Comments/Comment.js
+++ b/containers/Comments/Comment.js
@@ -1,4 +1,4 @@
-import React, { useState, useCallback } from 'react';
+import React, { useState, useCallback, useEffect } from 'react';
 import { View, Image } from 'react-native';
 import { Button } from 'react-native-elements';
 import { styles } from './Comment.styles';
@@ -24,8 +24,14 @@ export const Comment = (props) => {
     ratings: {
       negative_ratings:downVotes,
       positive_ratings:upVotes,
-    }
+    },
+    id: pk,
+    comment_content_type: commentContentType,
   } = props.comment;
+
+  useEffect(() => {
+    if (props.openSubComments) {setShowSubComments(true);}
+  }, [props.openSubComments]);
 
   const toggleSubComments = () => {
     setShowSubComments(!showSubComments);
@@ -105,6 +111,7 @@ export const Comment = (props) => {
           titleStyle={styles.buttonTitle}
           type='clear'
           styles={styles.commentButton}
+          onPress={() => {props.handleReply(commentContentType, pk);}}
         />
         <Button
           icon={redoIcon}
--- a/containers/Comments/CommentForm.js
+++ b/containers/Comments/CommentForm.js
@@ -1,8 +1,12 @@
 import React from 'react';
+import { View, TouchableOpacity } from 'react-native';
 import { Formik } from 'formik';
 import * as yup from 'yup';
 import { TextInputFormField } from '../../components/formFields';
-import { ButtonSubmit } from '../../components/ButtonSubmit';
+import { styles } from './CommentForm.styles';
+import IconSLI from 'react-native-vector-icons/SimpleLineIcons';
+import { COLORS } from '../../theme/colors';
+import { SIZES } from '../../theme/fonts';
 
 export const CommentForm = (props) => {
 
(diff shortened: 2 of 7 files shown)


containers/Comments/CommentForm.js
```diff
@@ -0,0 +1,54 @@
...
+        <>
+          <TextInputFormField
+            name='comment'
+            value={values.comment}
+            placeholder='Enter your comment'
+            onChangeText={handleChange('comment')}
+            onBlur={handleBlur('comment')}
+            error={errors.comment}
+            touched={touched.comment}
+            autoFocus = {props.isFocused}
```
Rineee, line 42: @khamui This is what doesnt work, so the form does not get focused when pops.isFocused is set to true. I also tried with extra state in CommentForm, that also did not help.

containers/Ideas/Idea.js
```diff
@@ -124,6 +128,43 @@ export const Idea = (props) => {
...
+          }
+          Alert.alert(errorMessage, errorDetail, [{ text: 'Ok' }]);
+        }
+      });
+
+  };
+
+  const handleCommentReply = (commentContentType, commentObjectPk) => {
+    setObjectToComment({'contentType': commentContentType, 'pk': commentObjectPk});
+    setFocusCommentForm(true);
```
Rineee, line 165: @khamui funny thing is that when you intialize FocusCommentForm with true and then set to false here, it looses focus when clicking on reply, so the other way around it works :shrug:

containers/Comments/Comment.js
```diff
@@ -24,9 +24,15 @@ export const Comment = (props) => {
...
       negative_ratings:downVotes,
       positive_ratings:upVotes,
-    }
+    },
+    id: pk,
+    comment_content_type: commentContentType,
   } = props.comment;
 
+  useEffect(() => {
+    if (props.openSubComments) {setShowSubComments(true);}
```
khamui, line 33: suggestion: in case you want to get rid of the `if` and some `{}`, think `props.openSubComments && setShowSubComments(true);` would work too.

containers/Comments/CommentForm.js
khamui: i pushed something very drafty (i tried to make it nice, but it was not as easy as i thought in the beginning though).

What could be the problem that `isFocused` does not work here is that `CommentForm` has been rendered already and while it is still `living` the change of this focus variable does not have an effect on it.

In my commit I make it dependable on a variable called `showCommentForm` which controls the entire component to be rendered or not. This way a change of that variable causes a new render of the component and therefore the variable applies.

Note: I am not sure if this behavior is what we want or not. Maybe we can have a quick talk tomorrow? Maybe @phillimorland wants to join as well?

containers/Ideas/Idea.js
```diff
@@ -20,6 +21,9 @@ export const Idea = (props) => {
   const [deleteModalVisible, setDeleteModalVisible] = useState(false);
   const [processing, setProcessing] = useState(false);
   const [comments, setComments] = useState([]);
+  const [objectToComment, setObjectToComment] = useState({'contentType': idea.content_type, 'pk': idea.pk});
```
khamui, line 24: i was a little confused by the name of this state variable. Not sure of a better alternative though?
Maybe `commentMeta` or something?

containers/Ideas/Idea.js
khamui: I didn't manage to make the reply request work. it throws an error.
Somehow the contenttype is undefined:
Bildschirmfoto 2022-01-25 um 16 13 17

containers/Comments/CommentForm.js
Rineee: @khamui Cool, the focusing works! Although I discussed the functionality a bit different with luca: the comment form should always be there at the bottom (without having to click on add comment) and when the user clicks on reply it gets focused (thats why I named it commentFormFocused).. but we can also discuss it again with him?
With the current version (probably also before), there is a problem when you clicked on add comment, then dont add one and click on reply -> nothing happens ?

containers/Comments/Comment.js
Rineee: To be honest, I actually prefer the more explicit notation :) But since we use the short notation everywhere, I will update it also!

containers/Ideas/Idea.js
Rineee: Yeah, totally also not happy with this name! It holds the content_type and pk of the object that is currently commented on, so either the idea or a comment... maybe commentOn? What do you think @khamui?

Rineee:
@khamui I just wanted to change the notation with the if statement, but then rebasing gets really annoying in the other PR, that is dependent on this one.. would it be fine for you if merge this and then go on from there? I will add a checklist above, so I do not forget..

containers/Comments/Comment.js
khamui: then lets go with your preferred solution! it was just a suggestion :)

containers/Comments/CommentForm.js
khamui: i see. actually my drafty version is just an attempt and i agree on the user flow it is a bit weird.
Currently the focus only works if the CommentForm is re-rendered, either clicking on the SubmitButton or the ReplyButton. A `"trigger only by focus"` solution we would need to research and tryout a little bit i guess.

Yes lets have a chat about!

containers/Ideas/Idea.js
khamui: hmm..yes could be. Or something like `commentContext`, `parentOfComment`...uff i am bad at this, looking at `objectToComment` makes kind of sense now..😬

or why not `contentTypeOfComment`?

khamui:
really nice work! merged to continue with #233

containers/Ideas/Idea.js
Rineee: @khamui What do you think of contentObjectOfComment? That would make sense in the way, that in django, the comment model does have a [content_object](https://github.com/liqd/adhocracy4/blob/db606bfb85f5af3e6c0cb13b1ce723adba681091/adhocracy4/comments/models.py#L22) that hold the content_type and pk of the commented object. So actually just what we have here..

Issue #155, opened 2021-09-23
idea detail view: three dots and back button incomplete


lthueer:
**URL:** 
**user:** 
**expected behaviour:** 
**behaviour:** three dots disappear and back button is cut off slightly 
**important screensize:**
**device & browser:** APPLE iPad mini (5. Generation, 2019) Wi-Fi, Tablet,
64 GB, 7,9 Zoll, Space Grey
**Comment/Question:** 

![20210923_171128](https://user-images.githubusercontent.com/36957474/134537018-0cf90232-6baf-4159-89af-2dc9262c3a5e.jpg)

phillimorland:
should be fixed

PR #226, opened 2022-01-18
[#5171] description textinput view

phillimorland:
This is just the styling, no form submission - should not be merged yet

### DIFF
--- a/containers/Ideas/Idea.styles.js
+++ b/containers/Ideas/Idea.styles.js
@@ -12,11 +12,18 @@ export const styles = StyleSheet.create({
     paddingHorizontal: SPACINGS.multiplyBy(.75),
     backgroundColor: COLORS.paper.main,
   },
-  contentContainer: {
-    paddingVertical: SPACINGS.base
+  descripContainer: {
+    paddingHorizontal: SPACINGS.multiplyBy(.75),
+    backgroundColor: COLORS.paper.main,
+    flex: 1,
+  },
+  descripBtnContainer: {
+    marginHorizontal: SPACINGS.multiplyBy(-.75),
+    flex: 1,
+    justifyContent: 'flex-end'
   },
   actionsContainer: {
-    height: SPACINGS.multiplyBy(2),
+    height: SPACINGS.multiplyBy(3),
     flexDirection: 'row',
     justifyContent: 'space-between',
     alignItems: 'center',
--- a/containers/Ideas/IdeaCreate.js
+++ b/containers/Ideas/IdeaCreate.js
@@ -22,7 +22,7 @@ import API from '../../BaseApi';
 
 export const IdeaCreate = props => {
 
-  const {idea, project, editing} = props.route.params;
+  const {idea, project, editing, descriptionText } = props.route.params;
   const moduleId = project.single_agenda_setting_module;
   const arrowLeftIcon = (
     <IconSLI name='arrow-left' size={22} />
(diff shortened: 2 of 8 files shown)


containers/Ideas/IdeaCreate.js
```diff
@@ -38,8 +39,7 @@ export const IdeaCreate = props => {
       .required('Idea title is Required'),
     description: yup
       .string()
-      .min(20, ({ min }) => `Description must be at least ${min} characters`)
-      .max(100, 'Description must be no longer then 100 characters')
+      .max(10000, 'Description must be no longer then 100 characters')
```
khamui, line 42: I guess this is still needed in this view to show, validation error while not in the description component?
10000 feels quite a lot of chars though.

just saw the text says 100, but the actual limit is 10000

containers/Ideas/IdeaCreate.js
```diff
@@ -161,6 +161,11 @@ export const IdeaCreate = props => {
 
   const onSetImage = (image) => setImage(image);
 
+  const pressHandler = () =>
```
khamui, line 164: it might be good to have a more specific name for this handler (since we are in a container component)

containers/Ideas/IdeaCreateDescription.js
```diff
@@ -0,0 +1,146 @@
...
+  const ideaDescriptionValidationSchema = yup.object().shape({
+    description: yup
+      .string()
+      .max(10000, 'Description must be no longer then 100 characters')
+      .required('Description is required'),
+  });
+
+  const [error, setError] = useState();
+
+  const handleSubmit = (values) => {
```
khamui, line 31: do we want to actual submit the description text separately from the IdeaCreate screen?
Or should we rather send it back to IdeaCreate and submit it to the backend together with the other fields?

containers/Ideas/IdeaCreate.js
phillimorland: this i left there till we knew what limit should be but there actually no limit so will remove

containers/Ideas/IdeaCreateDescription.js
phillimorland: I don't know, I kind of thought it would have to be submitted  for some reason but I guess the textinput value can just be sent back to the form in props and submitted as part of the IdeaCreate form right?

containers/Ideas/Idea.styles.js
```diff
@@ -12,11 +12,18 @@ export const styles = StyleSheet.create({
...
+  },
+  descripBtnContainer: {
+    marginHorizontal: SPACINGS.multiplyBy(-.75),
+    flex: 1,
+    justifyContent: 'flex-end'
   },
   actionsContainer: {
-    height: SPACINGS.multiplyBy(2),
+    height: SPACINGS.multiplyBy(3),
     flexDirection: 'row',
```
phillimorland, line 27: this fixes #155

containers/Ideas/IdeaCreateDescription.js
khamui: okay, yes i know this is part2 of the work. But so you also prefer to send it back and rather send it together with the rest right? Then leave it like that and we will see once we work on part2.

phillimorland: yes I think that would be best as the title of the toDescription button will need to change when description is added so the form will need to know it

khamui:
@phillimorland i added the form submission logic on your branch here, because i find it easier to merge it all together.

So your part is approved.
I am asking you to review the last commit about form submission :). If that works, feel free to merge.

containers/Ideas/IdeaCreateDescription.js
```diff
@@ -26,50 +23,23 @@ export const IdeaCreateDescription = props => {
       .required('Description is required'),
   });
 
-  const [error, setError] = useState();
+  const handleSubmitNoSave = (values) => {
```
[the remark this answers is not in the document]
phillimorland: i've done it as needed to adjust my code anyway :)

phillimorland (commented):
Awesome!

PR #227, opened 2022-01-18
auto-select api url

goapunk:
based on
* device (emulator/web/real)
* build type (PROD/DEV)
* environment variable LOCAL_API


khamui (commented):
currently i am starting expo with `expo start -c` which clears the Metro Bundler Cache.
i thought it helped in some cases.

is there a way to include that somehow?

goapunk:
hmm, we can either have commands for with or without cache and local or remote but that will increase the number of commands quite a lot. Alternatively we could use an environment variable, e.g. ``export LOCAL_API=true`` which you have to set once in the terminal and it will then work for all commands. Any preference?

khamui:
cool the solution you suggest setting LOCAL_API works fine!

PR #228, opened 2022-01-18
ideas/ideaCreate: preventing re-rendering while interacting with form

khamui:
render cycle caused the form to be re-rendered and therefore user inputs got
vanished.

### DIFF
--- a/components/formFields.js
+++ b/components/formFields.js
@@ -4,7 +4,6 @@ import { CheckBox } from 'react-native-elements';
 import DropDownPicker from 'react-native-dropdown-picker';
 import { styles } from './formFields.styles';
 import { TextSourceSans } from './TextSourceSans';
-import IconFA from 'react-native-vector-icons/FontAwesome';
 import { COLORS } from '../theme/colors';
 
 /*  Setting list mode to scrollview globally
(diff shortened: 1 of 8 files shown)


components/formFields.js
```diff
@@ -38,34 +38,82 @@ export const CheckBoxFormField = (props) => {
...
       {props.children.touched && <TextSourceSans style={styles.formError}>{props.error}</TextSourceSans>}
     </View>
   );
 };
 
-export const CustomCheckBoxFormField = (props) => {
-  const plusIcon = (
-    <IconFA name='plus' size={16} color={COLORS.text.main} />
-  );
+export const LabelList = props => {
```
phillimorland, line 51: I think this can go into it's own component as it's now so specific, I started a general form field file for reuseable, basic form fields but I think this is now past that, what do you think?

components/formFields.styles.js
```diff
@@ -48,7 +48,7 @@ export const styles = StyleSheet.create({
     marginLeft: 0,
     borderRadius: BORDERRADIUS.lg,
   },
-  customCheckBoxContainerParent: {
+  customCheckBoxContainer: {
```
phillimorland, line 51: I'm not sure on the naming here as the above styles are technically the container styles, but you are right Parent is a bit confusing, maybe `customCheckBoxesContainer`? god I hate naming things :(

components/formFields.js
```diff
@@ -38,34 +38,82 @@ export const CheckBoxFormField = (props) => {
...
+    props.onIconPress(selectedLabelsCopy);
+  };
+
+  const getCheckState = (choice) => {
+    return !!props.selectedLabels.find(sl => sl.id === choice.id);
+  };
+
+  return props.labelChoices.map(choice => (
+    <LabelFormField
+      key={`customCheckBox_${choice.value}`}
```
phillimorland, line 74: this doesn't seem to work getting `Encountered two children with the same key, `customCheckBox_undefined`.` error

phillimorland (commented):
Form fields working perfectly and nice refactor, hopefully will make it easier in future. Just few comments. Also does this also complete the task in #5169 with the image scroll? it seems to but task isn't asigned or moved or was it an accidental fix?

components/formFields.styles.js
khamui: maybe now that it is containing the specific Label component, i might change it to something specific in its own file.

components/formFields.js
khamui: yes agree!

khamui:
> Form fields working perfectly and nice refactor, hopefully will make it easier in future. Just few comments. Also does this also complete the task in #5169 with the image scroll? it seems to but task isn't asigned or moved or was it an accidental fix?

oh, not explicitly tested, so accidental then.

khamui:
#226 will be integrated after this PR is approved and merged

phillimorland (approved):
very nice! :)

PR #230, opened 2022-01-19
api: include release-channels in api url selection

goapunk:
I forgot to add cases for the release-channels we publish to. This should fix the login again.
I also added an eslint rule which stops the complaining about the switch-case, I couldn't get it to accept the code otherwise with a sane indentation


Rineee (approved):
Dont really know how to test this, but looks good, so lets try :)

PR #233, opened 2022-01-25
Ks 2022 01 rate comments

Rineee:
So this is branched off of https://github.com/liqd/zt-app/pull/223 which needs to be merged first and then this needs rebase.

### DIFF
--- a/containers/Ideas/Idea.js
+++ b/containers/Ideas/Idea.js
@@ -92,25 +92,24 @@ export const Idea = (props) => {
     setMenuVisible(false)
   );
 
-  const handleVote = async(direction) => {
+  const handleRate = async(value) => {
     if (processing) return;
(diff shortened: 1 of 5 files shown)


containers/Ideas/Idea.js
```diff
@@ -92,25 +92,24 @@ export const Idea = (props) => {
     setMenuVisible(false)
   );
 
-  const handleVote = async(direction) => {
+  const handleRate = async(value) => {
```
Rineee, line 95: @khamui I renamed this, because in Aplus its called rating everywhere and also in the serialized data, so I thought this is more consistent.. I hope that is fine for you, sorry for messing with your code :grimacing:

phillimorland (approved):
Very nice!

PR #239, opened 2022-01-26
better error handling and pick image from library fixed

khamui:


phillimorland (commented):
currently getting following error when try to save an image `An unhandled error was caught from submitForm(), [TypeError: undefined is not an object (evaluating 'iter[Symbol.iterator]')]`

containers/Ideas/IdeaCreate.js
```diff
@@ -122,7 +122,15 @@ export const IdeaCreate = props => {
           });
         }
         else if (statusCode == 400) {
-          setError('Required fields are missing.');
+          if (data && data.image) {
+            const errorString = data.image.reduce((acc, curr) => {
+              return `${acc}\n${curr}`;
+            }, 'Please try again.\n');
```
phillimorland, line 128: is this error message meant to convay that the image is the wrong size? maybe check the error messages in a+

phillimorland: shown along with backend message

PR #244, opened 2022-01-27
WIP imageFormFIeld: add image rm button similar to styling but not exact

phillimorland:


Rineee:
@phillimorland Removing the image from the form works perfectly, but then when I submit the idea, it is still sent?

PR #273, opened 2022-06-23
richtext (ckeditor PoC)

khamui:
Note: needs backend to send `project.information` without striptags.

### DIFF
--- a/components/Richtext.js
+++ b/components/Richtext.js
@@ -1,0 +1,71 @@
+import React from 'react';
+import { useWindowDimensions } from 'react-native';
+import { baseUrl } from '../BaseApi';
+import { RichtextCollapsibleItem } from './RichtextCollapsibleItem';
+import RenderHTML, { useInternalRenderer } from 'react-native-render-html';
+
+// The <Richtext /> component can be used as a container
+// for HTML input. It converts html to React Native with the
+// help of a package called react-native-render-html.
+//
+// 1. <Richtext> receives a html formatted prop called "information"
+// 2. A Subcomponent <RenderRichtext> creates the dom and alters two things:
+//    a.) image sources: relative to absoulte url
+//    b.) collapsibles: converting div to article
+// 3. <RenderRichtext> uses <RenderHTML> to convert html into React Native code.
+// 4. While doing this it uses a custom renderer (collapsibleRenderer) and
+//    replaces all <article> elements with a custom component <RichtextCollapsibleItem>
+export const Richtext = ({ project: { information } }) => {
+  const {width} = useWindowDimensions();
+
+  const isCollapsible = (node) => node.domNode.attribs.class === 'collapsible-item';
+
+  const extractCollapsibleData = (tnode) => {
+    return {
+      title: tnode.children[0].children[0].children[0].data,
+      body: tnode.children[1].children[0].children[0].children[0].data
+    };
+  };
+
+  const collapsibleRenderer = ({ TDefaultRenderer, tnode,  ...props }) => {
+    if (isCollapsible(tnode)){
+      const { title, body } = extractCollapsibleData(tnode);
+      return (
+        <TDefaultRenderer
+          tnode={tnode}
+          {...props}
+        >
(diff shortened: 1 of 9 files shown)


khamui:
Needs backend to send html formatted text:
https://github.com/liqd/adhocracy-plus/pull/1953 

The third commit is a version where collapsibles are not working, but also is much simpler codewise (and less dependencies). All other formattings (bold, italic, links etc.) are working

The fourth commit is a version where collapsibles work, but is much more complex codewise and includes a proprietary dependency called `domutils`.

@fuzzylogic2000 @phillimorland @goapunk @Rineee lets discuss and decide what we should do about richtext in the app?

khamui:
Actually i am not sure if its proprietary, just stumbled over the text about `enterprise`? But not sure if this is counting in our case?
https://github.com/fb55/domutils#domutils-for-enterprise

goapunk:
yeah, it's some kind of service. The license itself is free (you need to somehow include the license texts though)

khamui:
> yeah, it's some kind of service. The license itself is free (you need to somehow include the license texts though)

okay, well for a discussion it is good to know that its not proprietary 👍🏽

khamui:
@Rineee @goapunk some screenshots to see the `collapsibles`:

with collapsibles:
Bildschirmfoto 2022-06-30 um 10 06 33
Bildschirmfoto 2022-06-30 um 10 06 43
Bildschirmfoto 2022-06-30 um 10 06 15


without collapsbiles:
Bildschirmfoto 2022-06-30 um 10 17 27
Bildschirmfoto 2022-06-30 um 10 17 34

goapunk:
thanks, it looks okay for me (not super hacky), so I'm in principle fine with the collapsible version. I've also found a way to make it work without domutils, if that's something we want to avoid: https://github.com/liqd/zt-app/commit/b053e04fca7f3c90006fef80601caf95956603fb

goapunk:
if we are going to merge this we can remove `domutils` from `package.json` and should probably also fix the eslint errors

khamui:
- [ ] removing random included richtext and put it into tabs (when tabs are available)

khamui:
can be reviewed (keep in mind it needs the aplus backend PR from above)

fuzzylogic2000:
@goapunk I don't want to tag in, but can we merge this? I mean, I don't want to tag into the reviewing when it looks like you are very much on it. But I can also take a look if need be.

goapunk:
I'd review it next but you can totally have a look (and merge) if you want

components/Richtext.js
```diff
@@ -0,0 +1,82 @@
...
+// for HTML input. It converts html to React Native with the
+// help of a package called react-native-render-html.
+//
+// 1. <Richtext> receives a html formatted prop called "text"
+// 2. A Subcomponent <RenderRichtext> creates the dom and alters two things:
+//    a.) image sources: relative to absoulte url
+//    b.) collapsibles: converting div to article
+// 3. <RenderRichtext> uses <RenderHTML> to convert html into React Native code.
+// 4. While doing this it uses a custom renderer (collapsibleRenderer) and
+//    replaces all <article> elements with a custom component <RichtextCollapsibleItem>
```
goapunk, line 18: we don't use the article anymore, right?

goapunk (approved):
Really cool! As this is quite the big thing I will assign @fuzzylogic2000 to also have a look

components/RichtextCollapsibleItem.js
```diff
@@ -7,16 +7,20 @@ import { styles } from './RichtextCollapsibleItem.styles';
 export const RichtextCollapsibleItem = ({ title, body }) => {
   const [collapsed, setCollapsed] = useState(true);
   const chevron = collapsed
```
fuzzylogic2000, line 9: chevron and plus/minus? Can we call it sth. more general? Maybe collapseItem? I cannot think of sth. that makes looots of sense, but it's not a chevron anymore. :)

fuzzylogic2000 (approved):
Nice, looks good!
I will merge and we can fix the smaller things in the next iteration of these tabs.

PR #295, opened 2022-06-29
making base api working with local_api and export baseUrl (without /api)

khamui:


BaseApi.js
```diff
@@ -1,41 +1,43 @@
...
 import * as Device from 'expo-device';
 
-const baseUrl = (() => {
+export const baseUrl = (() => {
   /*global __DEV__*/
   if (__DEV__) {
     if (Constants.manifest.extra.localAPI) {
-      if (Device.isDevice) {
-        return 'http://localhost:8004/api';
+      if (Device.isDevice && !Device.brand === 'google') {
```
goapunk, line 8: probably missed it but why do we need this?

goapunk: but why would you start it with `localAPI=True` on a real phone? I think I added this case for the web (which was detected as device). On a real device the url would have to be the one of your laptop running local aplus.

khamui: strange, because when running in the emulator isDevice is true too?

khamui: Yes sound like a better workaround, will change and test it! thanks for checking.

goapunk:
cool!

PR #299, opened 2022-06-29
Jd 2022 06 comments editable

goapunk:
* make header (back button and settings button) sticky when scrolling (so you can cancel the editing, ~~we don't seem to handle taps yet, tapping somewhere on the screen above the form should also cancel it, this is a TODO~~, same for hardware back button on android and swipe gesture on iOS?)
*  add Pressable which cancels editing mode if you press outside the input and applies the transparent grey overlay
* make settings button for Idea editing disappear on scrolling (not sure if this is what we want? The settings button wasn't in the design but that could have bee because it's not from admin/initiator view?)
* make comments editable

### DIFF
--- a/BaseApi.js
+++ b/BaseApi.js
@@ -177,6 +177,12 @@ const API = {
     const url = ct_url.replace('$objectPk', objectPk);
     return makePostRequest(url, data, token);
   },
+  editComment(contentTypeId, objectPk, commentPk, data, token = null) {
+    const ct_url = endpoints.comment.replace('$contentTypeId', contentTypeId);
+    const ct_obj_url = ct_url.replace('$objectPk', objectPk);
+    const url = ct_obj_url.replace('$commentPk', commentPk);
+    return makePutRequest(url, data, token);
+  },
   deleteComment(contentTypeId, objectPk, commentPk, token = null) {
     const ct_url = endpoints.comment.replace('$contentTypeId', contentTypeId);
     const ct_obj_url = ct_url.replace('$objectPk', objectPk);
--- a/containers/Comments/Comment.js
+++ b/containers/Comments/Comment.js
@@ -24,7 +24,7 @@ export const Comment = (props) => {
     {
       title: 'Edit',
       icon: 'pencil',
-      action: () => console.log('Edit comment'),
+      action: () => props.toggleEditing(commentBeingProcessed),
       isFirst: true,
       isAllowed: comment.user_info.has_changing_permission
     },
@@ -88,6 +88,18 @@ export const Comment = (props) => {
   const onTextLayout = useCallback(e => {
     setHasExcerpt(e.nativeEvent.lines.length > NUM_OF_LINES);
   }, []);
+
+  function handleOptions(subcomment) {
+    if (subcomment !== undefined){
+      setCommentBeingProcessed(subcomment);
+    }
+    else {
+      setCommentBeingProcessed(comment);
+    }
+    props.setDeleteModalItems(commentDeleteModalItems);
+    props.setMenuItems(commentMenuItems);
+    props.toggleMenu();
+  }
 
   const handleRate = async(ratingComment, value) => {
     if (processing) return;
@@ -198,7 +210,7 @@ export const Comment = (props) => {
           <Button
             icon={optionsIcon}
             type='clear'
-            onPress={() => {setCommentBeingProcessed(comment); props.toggleMenu();}}
+            onPress={() => handleOptions()}
           />
         </TextSourceSans>
         }
@@ -274,7 +286,7 @@ export const Comment = (props) => {
         <SubComments
           comments={comment.child_comments}
           handleRate={handleRate}
-          setCommentBeingProcessed={setCommentBeingProcessed}
+          handleOptions={handleOptions}
           toggleMenu={props.toggleMenu}
           getCommentTextDisplay={getCommentTextDisplay}
           isDisplayed={isDisplayed}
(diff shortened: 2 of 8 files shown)


goapunk:
wip because I'm not sure about the sticky scrolling behavior

containers/Ideas/Idea.js
```diff
@@ -23,6 +23,7 @@ export const Idea = (props) => {
   const [comments, setComments] = useState([]);
   const [contentObjectOfComment, setContentObjectOfComment] = useState({'contentType': idea.content_type, 'pk': idea.pk});
   const [commentLastCommented, setCommentLastCommented] = useState(-1);
+  const [isScrolling, setIsScrolling] = useState(false);
```
khamui, line 26: really cool with the sticky back button and also physical back works!
Just a question regarding the IdeaProject.js. Do we want to have a sticky backbutton there too? Maybe more of a question for @mcastro-lqd .

And if yes, would it be easy to implement this there as well? or shall we see that later?

BaseApi.js
```diff
@@ -173,6 +173,12 @@ const API = {
     const url = ct_url.replace('$objectPk', objectPk);
     return makePostRequest(url, data, token);
   },
+  editComment(contentTypeId, objectPk, commentPk, data, token = null) {
+    const ct_url = endpoints.comment.replace('$contentTypeId', contentTypeId);
```
khamui, line 177: i know it is the pythonic way, but to keep consistency here i would rather do camelcase notation?
eslint should catch this, though

containers/Comments/Comment.js
```diff
@@ -89,6 +89,18 @@ export const Comment = (props) => {
     setHasExcerpt(e.nativeEvent.lines.length > NUM_OF_LINES);
   }, []);
 
+  function handleOptions(subcomment) {
+    if (subcomment !== undefined){
```
khamui, line 93: can subcomment be null or 0? if not maybe just `if (!subcomment)` ?

containers/Comments/Comment.js
```diff
@@ -198,7 +210,7 @@ export const Comment = (props) => {
           <Button
             icon={optionsIcon}
             type='clear'
-            onPress={() => {setCommentBeingProcessed(comment); props.toggleMenu();}}
+            onPress={() => handleOptions()}
```
khamui, line 213: because no params passed, would `onPress={handleOptions}` work too?

BaseApi.js
```diff
@@ -173,6 +173,12 @@ const API = {
     const url = ct_url.replace('$objectPk', objectPk);
     return makePostRequest(url, data, token);
   },
+  editComment(contentTypeId, objectPk, commentPk, data, token = null) {
```
khamui, line 176: root level comment editing does not work for me:
Bildschirmfoto 2022-06-30 um 12 17 14

child level comment editing works fine!

containers/Ideas/Idea.js
```diff
@@ -320,76 +320,78 @@ export const Idea = (props) => {
...
-          )}
-          <TextSourceSans style={styles.text}>{ideaState.description}</TextSourceSans>
-        </View>
-        {getLabels().length > 0 && (
-          <View style={styles.labelsContainer}>
-            {getLabels().map((label, idx) => (
-              <Label key={idx + label} title={label} />
-            ))}
+        <Pressable onPress={isEditing && toggleEditing} style={{
+          ...isEditing ? {opacity: 0.25, backgroundColor: '#fff'}: {}
```
khamui, line 324: maybe we could use the same lightbox overlay (greyish) which we have when opening the three dots menu?

containers/Comments/Comment.js
goapunk: it's undefined if called from comment and defined if called from a subcomment, so I think we need the undefined check?

containers/Comments/Comment.js
goapunk: but then it passes the event from `onPress`, which clashes with the possible subcomment param of `handleOptions`

containers/Ideas/Idea.js
goapunk: oh, yes, didn't see that! will check

containers/Comments/Comment.js
khamui: yes in that case undefined

BaseApi.js
goapunk: I just copied this from the other api functions :) so apparently eslint didn't complain before

BaseApi.js
goapunk: fixed, thanks!

containers/Ideas/Idea.js
goapunk: hmm, we are using a bottomsheet for that, I'm not sure how to integrate that nicely (it's also more grey than in the design I think?).

containers/Ideas/Idea.js
khamui: lets leave the back button on IdeaProject out for now and talk about it later

BaseApi.js
khamui: i see, okay lets leave it for now

containers/Comments/Comment.js
khamui: i don't follow, the notation i suggestion would call a different function that has the same name?

containers/Ideas/Idea.js
khamui: i see, maybe at least we can change it to the design grey, instead of white? And see if we can adjust the grey of the bottomsheet?

containers/Comments/Comment.js
goapunk: yes, handleOptions can be called from both comment and subcomment. If called from the comment we already have all the info we need, if called from the subcomment we need to tell it which subcomemnt. I think the difference between this and yours is: `onPress={handleOptions}` is equivalent to `onPress={(event)=>handleOptions(event)}`. Since the first param of `handleOptions` is supposed to be `subcomment` we would end up with the event being passed as the subcomment

containers/Ideas/Idea.js
goapunk: don't worry :) also if you know a good way to solve it with the bottomsheet feel free to push it here :)

PR #306, opened 2022-07-04
Jd 2022 07 comments test

goapunk:
* fix some unncessary useStates
* add some tests for comments + ideas

### DIFF
--- a/containers/Comments/Comment.js
+++ b/containers/Comments/Comment.js
@@ -207,6 +207,7 @@ export const Comment = (props) => {
         {isDisplayed(comment) &&
         <TextSourceSans>
           <Button
+            testID={'options_button_' + comment.id}
             icon={optionsIcon}
             type='clear'
             onPress={() => handleOptions()}
--- a/containers/Comments/Comment.js
+++ b/containers/Comments/Comment.js
@@ -1,4 +1,4 @@
-import React, { useState, useCallback, useEffect } from 'react';
+import React, { useState, useCallback, useEffect, useRef } from 'react';
 import { Alert, View, Image } from 'react-native';
 import { Button } from '@rneui/base';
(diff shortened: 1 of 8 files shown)


khamui (changes requested):
Pretty cool! thanks for the render optimizations. And the huge test coverage increase :)

Just one thing that prevents from rating:

containers/Comments/Comment.js
```diff
@@ -102,10 +97,12 @@ export const Comment = (props) => {
 
   const handleRate = async(ratingComment, value) => {
     if (processing) return;
-    setProcessing(true);
+    processing.current = true;
     const token = await AsyncStorage.getItem('authToken');
     const newComment = await rate(ratingComment, value, token);
-    newComment && setProcessing(false);
+    if(newComment) {
+      processing.current = true;
```
goapunk, line 104: this should be false :facepalm:

goapunk:
Should be fixed now, thanks for catching this!

phillimorland (approved):
looks good but will leave final check to @khamui as he started

containers/Comments/Comment.js
```diff
@@ -102,10 +97,12 @@ export const Comment = (props) => {
 
   const handleRate = async(ratingComment, value) => {
     if (processing) return;
```
khamui, line 99: still not possible to rate comments. Think `if (processing) return;` should be `if (processing.current) return;`?
empty objects are truthy, so in this case it will always early return.

khamui (changes requested):

containers/Comments/Comment.js
goapunk: arg, nice catch! fixed

khamui:
sweet! working fine

PR #307, opened 2022-07-05
BaseApi.js: in case of apple device (physical and simulator) use loca…

khamui:
…lhost


goapunk (approved):
Can't test as I don't have an Iphone but looks good to me!

PR #318, opened 2022-07-07
expo: make web version work again

goapunk:
helps with checking dimensions, margin, etc.


phillimorland (commented):
~~Opens the webversion however when attempting to login after adding credentials it freezes, I am not sure if this is due to the keyboard encoding of the @ being different but I tried copying and pasting my email from the emulator, which worked previously, however this seems to still not work, should I try testing in another way?~~
NOTE: only works with `make start-web` cannot use terminal to open web interface

phillimorland (approved):
cool!

PR #323, opened 2022-07-09
[#6178] IdeaProject: make tab/buttons show correct content when clicked

phillimorland:
after previous attempts with tab libs was decided using buttons would be simpler


phillimorland:
@goapunk I have sorted the gradient and active styling, the grey area when little to no content I poked at but will be a bit longer and I don't want to waste time before I know how we will get the content in there so will add a task to that story as thats kind of where the styling for that should go anyway, what do you think?

goapunk (approved):
Works nicely! I added a commit which only uses a single state instead of 3. Not sure how much better that is but it somehow feels a bit easier to read and might save us some headache with react

phillimorland:
@goapunk cool, so can this be merged then?

PR #325, opened 2022-07-12
deps: update packages with expo command (disabling renovate)

khamui:


renovate.json
```diff
@@ -1,4 +1,5 @@
 {
+  "enabled": false,
```
fuzzylogic2000, line 2: This file should be removed!

package.json
```diff
@@ -47,18 +47,18 @@
     "expo-permissions": "13.2.0",
     "expo-splash-screen": "0.15.1",
     "expo-status-bar": "1.3.0",
-    "expo-updates": "0.13.2",
+    "expo-updates": "~0.13.3",
```
fuzzylogic2000, line 50: renovate does always pin the dependencies instead of using the ` ~`. I don't know the advantages and disadvantages of either. I guess like this is also fine?

fuzzylogic2000 (changes requested):
Nice, but the whole renovate file can now be removed.

package.json
khamui: yes we should remove `~` and `^` because we use expo's updating compatible versions which needs exact version numbers.

goapunk (approved):
looks good to me, nice change!

PR #326, opened 2022-07-12
containers/Auth: also show field errors

fuzzylogic2000:
We did actually already show the only error, I could find that could appear in the app. We do make sure in the app that we don't send an empty username or password. I still added the errors for the two fields we have, in case there are other errors that the user would otherwise miss. 
If you want to test password or username errors, you need to remove lines 34-39 and the `disabled={!isValid}` from the button.


khamui:
nice! removed the `yup` part to test.

PR #329, opened 2022-07-12
[#6145] adding report functionality to ideas 

phillimorland:


containers/Reports/ReportCreateMessage.js
```diff
@@ -0,0 +1,124 @@
...
+    <IconSLI name='arrow-left' size={22} />
+  );
+  const reportMessageValidationSchema = yup.object().shape({
+    message: yup
+      .string()
+      .max(1024, 'Message must be no longer then 1024 characters')
+  });
+
+  const handleSubmit = (values) => {
+    let description;
```
goapunk, line 28: this can be a const as it doesn't change if you do `const description = values.message` instead of two lines

BaseApi.js
```diff
@@ -144,6 +145,9 @@ const API = {
     const url = module_url.replace('$ideaPk', ideaPk);
     return makePutRequest(url, data, token);
   },
+  postReport(content_type, object_pk, description, token = null) {
```
goapunk, line 148: maybe replace the curly braces just with `data`  here and add the curly braces in the `handleSubmit` so it's more in style with the other functions here?

goapunk (approved):
Very nice!

goapunk (changes requested):
on a small screen the keyboard is overlaying the input and you can't see what you are typing.

phillimorland:
@goapunk can you specify test device used?

PR #330, opened 2022-07-12
WIP: container/login: add text with links and style

fuzzylogic2000:


fuzzylogic2000:
Oh, I forgot to ask: how do we do sizes in the app? I just did some guessing based on the font-sizes and spacing we use elsewhere in the app to make it kind of look like the design.
But is there a way to test and tell the pixels and do we normally do that?

khamui:
> 

tbh thats what i did until now (guessing and approximating). But the better way is to run in the browser and use devtools to measure it (if we want to make it pixel perfect).

containers/Auth/LoginScreen.styles.js
```diff
@@ -0,0 +1,29 @@
...
+  },
+  container: {
+    paddingHorizontal: SPACINGS.multiplyBy(.75),
+  },
+  title: {
+    marginTop: SPACINGS.multiplyBy(4),
+    marginBottom: SPACINGS.multiplyBy(0.75),
+    fontSize: SIZES.xl,
+    fontWeight: FONTWEIGHT.boldxl
+  },
```
phillimorland, line 20: it would be good if all of these could be defined once and just imported but that's for another day I think, along with the font clean ups

components/LinkTextSourceSans.js
```diff
@@ -0,0 +1,20 @@
+import React from 'react';
+import { StyleSheet } from 'react-native';
+import { TextSourceSans } from './TextSourceSans';
+
+export const LinkTextSourceSans = (props) => {
+  const mergedStyles = props.style
+    ? [styles.font, props.style]
+    : styles.font;
+  return (
+    <TextSourceSans {...props} style={mergedStyles}>
```
phillimorland, line 10: very nice!

phillimorland (commented):
very nice, except keyboard covers the password input on smaller screens, tested on nexus 4, although possible it always has?

fuzzylogic2000:
> very nice, except keyboard covers the password input on smaller screens, tested on nexus 4, although possible it always has?

@phillimorland That probably came with the more text at the top screen. What can I do about it? Make the upper part scrollable?

phillimorland:
@fuzzylogic2000 I think if you wrap a `KeyboadAvoidView` in a `ScrollView` then it should work, however flex doesn't play nice, i was looking at a way to do it with the description but in the end a large text area worked but maybe I look at it again. If we wrap the whole think in a `View` with flex it might work?!

containers/Auth/LoginScreen.styles.js
fuzzylogic2000: Yeah, we do still have a bit of a mess with the styles. :woman_shrugging:

containers/Auth/LoginScreen.js
```diff
@@ -65,45 +69,55 @@ export const LoginScreen = () => {
...
-            onChangeText={handleChange('password')}
-            onBlur={handleBlur('password')}
-            error={errors.password}
-            touched={touched.password}
-            secureTextEntry
-          />
-          <Button onPress={handleSubmit} title="Login" disabled={!isValid} />
+        <View style={styles.flexContainer}>
+          <View style={styles.container}>
+            <ScrollView>
```
fuzzylogic2000, line 74: @phillimorland It was really that easy! Just adding the ScrollView here worked! :)

phillimorland (approved):
perfect! :)

PR #331, opened 2022-07-14
[#6146] report ideas and comments

phillimorland:
branched off of https://github.com/liqd/zt-app/pull/329 and continued working on this one, kind of by accident but felt it was bit late once I realised

### DIFF
--- a/BaseApi.js
+++ b/BaseApi.js
@@ -44,6 +44,7 @@ const endpoints = {
   rating: baseApiUrl + '/contenttypes/$contentTypeId/objects/$objectPk/ratings/$ratingId/',
   comments: baseApiUrl + '/contenttypes/$contentTypeId/objects/$objectPk/comments/',
   comment: baseApiUrl + '/contenttypes/$contentTypeId/objects/$objectPk/comments/$commentPk/',
+  report: baseApiUrl + '/reports/',
 };
 
 const getHeaders = (token, isFormData = false) => {
@@ -143,6 +144,9 @@ const API = {
     const module_url = endpoints.idea.replace('$moduleId', moduleId);
     const url = module_url.replace('$ideaPk', ideaPk);
     return makePutRequest(url, data, token);
+  },
+  postReport(data, token = null) {
+    return makePostRequest(endpoints.report, data, token);
   },
   postLogin(data) {
     return makePostRequest(endpoints.login, data);
--- a/containers/Comments/Comment.js
+++ b/containers/Comments/Comment.js
@@ -36,7 +36,10 @@ export const Comment = (props) => {
     {
       title: 'Report',
       icon: 'flag',
-      action: () => console.log('Report comment'),
+      action: () => {
+        props.toggleMenu();
+        props.navigation.navigate('ReportCreateMessage', {content_type: comment.comment_content_type, object_pk: comment.id});
+      },
       isFirst: !comment.user_info.has_changing_permission && !comment.user_info.has_deleting_permission,
       isLast: true,
       isAllowed: true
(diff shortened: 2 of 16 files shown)


containers/Comments/Comment.js
```diff
@@ -36,7 +36,10 @@ export const Comment = (props) => {
     {
       title: 'Report',
       icon: 'flag',
-      action: () => console.log('Report comment'),
+      action: () => {
+        props.toggleMenu();
+        props.navigation.navigate('ReportCreateMessage', {content_type: comment.content_type, object_pk: comment.object_pk});
+      },
       isFirst: !comment.user_info.has_changing_permission && !comment.user_info.has_deleting_permission,
```
Rineee, line 43: I think here you need {content_type: comment.comment_content_type, object_pk: id}, because content_type and object_pk are referring the the content object the comment belongs to, so here you are posting the report on the idea etc and not on the comment.

containers/Comments/Comment.js
```diff
@@ -36,7 +36,10 @@ export const Comment = (props) => {
     {
       title: 'Report',
       icon: 'flag',
-      action: () => console.log('Report comment'),
+      action: () => {
+        props.toggleMenu();
+        props.navigation.navigate('ReportCreateMessage', {content_type: comment.content_type, object_pk: comment.object_pk});
+      },
```
Rineee, line 42: I think here you need {content_type: comment.comment_content_type, object_pk: id}, because content_type and object_pk are referring the the content object the comment belongs to, so here you are posting the report on the idea etc and not on the comment.

BaseApi.js
```diff
@@ -144,6 +145,9 @@ const API = {
     const url = module_url.replace('$ideaPk', ideaPk);
     return makePutRequest(url, data, token);
   },
+  postReport(content_type, object_pk, description, token = null) {
+    return makePostRequest(endpoints.report, {content_type, object_pk, description}, token);
```
Rineee, line 149: Maybe you could pass data as a dict here and already handle that in handleSubmit? So I mean here just 
```postReport(data, token=null)``` and makePostRequest(endpoint, data, token) and in handleSubmit API.postReport({content_type, object_pk, description}, token).

Comment (phillimorland) [reply BaseApi.js]: thanks for explaining! I didn't get the difference in the requests before

**Comment by Rineee** [comment]
Wonderful, all working and good to merge from my side!

File: containers/Reports/ReportCreateMessage.js [hunk 4]
```diff
@@ -52,68 +52,66 @@ export const ReportCreateMessage = props => {
...
-              placeholder='Add message'
-              returnKeyType='next'
-              returnKeyLabel='next'
-              onChangeText={handleChange('message')}
-              onBlur={handleBlur('message')}
-              error={errors.message}
-              touched={touched.message}
-              numberOfLines={10}
-              multiline={true}
+    <View style={styles.container}>
```
khamui, line 55: the keyboard open/hide seems broken for Idea description and Report Modal.

android, google pixel 5
- clicking opens the keyboard, but overlaps the submit button (its possible to scroll though).
- hiding the keyboard again, the button is in the middle of the screen instead of the bottom of the screen

phillimorland: @khamui the keyboard overlapping submit is expected or on small screens you can only see a line or so of what you are typing. 
The submit button in middle is new, will test on pixel 5 and see if can replicate.

phillimorland: @khamui ok when fixing button position the overlapping also got fixed for smaller screen, hopefully

khamui:
yeah! working fine now!

PR #333, opened 2022-07-14
termsofuse/comments: setting terms of use agreement when commenting

khamui:


BaseApi.js
```diff
@@ -69,6 +69,7 @@ const makeGetRequest = (url, token = null) => {
 
 const makePostRequest = (url, data = {}, token = null) => {
   const isFormData = data instanceof FormData;
+  !isFormData && (data.agreed_terms_of_use = true);
```
fuzzylogic2000, line 72: That does work. But why? What does `isFormData` mean? Why don't we always add `agreed_terms_of_use` to the data?

goapunk: as a side note: doing it here technically means that if you revoke your terms of use through the web you won't be able to edit or delete comments because they use PUT and DELETE. Do we care about that?

Rineee: But shouldn't this go into the addComment/editComment below? Otherwise we are now also sending this when posting a rating or logging in.

fuzzylogic2000: Yes, I agree. It should be somewhere more self-explanatory and only on the comment. And we need to send it with create and update.
For the delete to do that without should be fine, as the whole agreement is more about creating content, right?

khamui: okay, i thought doing it for all POST request would avoid inconsistencies. but yeah, it is only for comments now (and polls, but for the app only comments), so will change it to only when creating and editing comments and not sending it for comment deletion.

fuzzylogic2000:
@khamui Is this ready to be reviewed again?

khamui:
@fuzzylogic2000 yes

fuzzylogic2000 (approved):
very nice!

PR #334, opened 2022-07-14
add header component, make header look coherent on all pages

goapunk:
fix the issue with the back button not being clickable if it's too close the the android status bar. Apparently I'm the only one having the issue though. This makes the position also more consistent with the Idealist view.

### DIFF
--- a/components/Header.styles.js
+++ b/components/Header.styles.js
@@ -8,7 +8,8 @@ export const styles = StyleSheet.create({
     backgroundColor: COLORS.paper.main,
   },
   actionsContainer: {
-    marginBottom: SPACINGS.multiplyBy(1),
+    paddingTop: SPACINGS.multiplyBy(0.5),
+    paddingBottom: SPACINGS.multiplyBy(1),
     flexDirection: 'row',
     justifyContent: 'space-between',
     alignItems: 'center',
--- a/components/Header.styles.js
+++ b/components/Header.styles.js
@@ -1,0 +1,25 @@
+import { StyleSheet } from 'react-native';
+import { COLORS } from '../theme/colors';
+import { SIZES } from '../theme/fonts';
+import { SPACINGS } from '../theme/spacings';
+
+export const styles = StyleSheet.create({
+  container: {
+    backgroundColor: COLORS.paper.main,
+  },
+  actionsContainer: {
+    marginBottom: SPACINGS.multiplyBy(1),
+    flexDirection: 'row',
+    justifyContent: 'space-between',
+    alignItems: 'center',
+  },
+  backButton: {
+    padding: 0,
+    marginVertical: 0,
+  },
+  backButtonText: {
+    fontSize: SIZES.md,
+    marginHorizontal: SPACINGS.multiplyBy(.25),
+    color: COLORS.text.main,
+  }
+});
--- a/containers/Ideas/Idea.js
+++ b/containers/Ideas/Idea.js
@@ -298,108 +298,110 @@ export const Idea = (props) => {
         />);
 
   return (
-    <KeyboardAvoidingView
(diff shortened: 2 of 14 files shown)


containers/Ideas/Idea.js
```diff
@@ -296,7 +296,7 @@ export const Idea = (props) => {
       behavior={(Platform.OS === 'ios')? 'padding' : null}
       style={{flex:1}}
     >
-      <View style={styles.container}>
+      <View style={styles.headerContainer}>
```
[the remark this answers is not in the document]
goapunk: yes, it's still used below for the main content

goapunk:
> The projects and ideas have the back back button quite far down now. But for idea edit it is behind the bar for me as well now. I didn't check how it behaved before, though.

The project one always has been that far down, so it's the same now for the idea. I agree that it looks...wrong? Also it seems to be more of an emulator problem. I don't have it on my real phone

goapunk:
I'll rework it, maybe it would also make sense to put that into it's own component?

containers/Ideas/Idea.styles.js
```diff
@@ -21,8 +21,12 @@ export const styles = StyleSheet.create({
     flex: 1,
     justifyContent: 'flex-end'
   },
+  headerContainer: {
```
khamui, line 24: is there a way to check the statusbar and set the height dynamically?

in both it looks quite far in the middle and does not match the design. I am attaching the iphone screenshot, for the android it is slightly closer to the statusbar but still quite for imho.

Bildschirmfoto 2022-07-14 um 17 57 04

khamui: ah sorry, just saw the WIP now...

phillimorland (commented):
just one little thing on consistency but otherwise good to go

containers/Reports/ReportCreateMessage.styles.js
```diff
@@ -6,24 +6,9 @@ import { SPACINGS } from '../../theme/spacings';
 export const styles = StyleSheet.create({
   container: {
     backgroundColor: COLORS.paper.main,
+    paddingTop: SPACINGS.multiplyBy(.25),
```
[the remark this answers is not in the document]
goapunk: somehow it was needed in `ideaCreate` and I didn't find out why it behaves different there. I just assumed it's because of some CSS stuff I don't get (padding of another element or dunno). If you know where to change I'm all for it.

components/Header.styles.js
```diff
@@ -0,0 +1,26 @@
+import { StyleSheet } from 'react-native';
+import { COLORS } from '../theme/colors';
+import { SIZES } from '../theme/fonts';
+import { SPACINGS } from '../theme/spacings';
+
+export const styles = StyleSheet.create({
+  container: {
+    backgroundColor: COLORS.paper.main,
```
goapunk, line 8: on web this doesn't seem to get inherited by actionsContainer so the background is grey, on android it works. Do we need to care?

containers/Reports/ReportCreateMessage.styles.js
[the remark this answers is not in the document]
goapunk: I moved the Header out of the KeyboardAvoidingView (as it apparently comes with padding/margin already) in `Idea.js` and set the padding on the Header directly now.

components/Header.styles.js
phillimorland: nope :)

goapunk: or maybe you can check if it works on iOS as well?

phillimorland: will do a check with the iphone but webview is often pretty far out

phillimorland: hmm currently getting no projects loading on any branch so will merge and test on dev

PR #346, opened 2022-07-19
Ks 2022 07 issues

Rineee:
Needs https://github.com/liqd/adhocracy-plus/pull/1990

### DIFF
--- a/containers/Ideas/IdeaProject.js
+++ b/containers/Ideas/IdeaProject.js
@@ -45,9 +45,9 @@ export const IdeaProject = (props) => {
   );
 
   const fetchIdeas = () => {
-    project.single_agenda_setting_module &&
+    project.single_idea_collection_module &&
       AsyncStorage.getItem('authToken')
-        .then((token) => API.getIdeas(project.single_agenda_setting_module, token))
+        .then((token) => API.getIdeas(project.single_idea_collection_module, token))
         .then((ideaResponse) => {
           setIdeas(ideaResponse);
         });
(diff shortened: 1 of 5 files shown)


containers/Ideas/Phase.js
```diff
@@ -0,0 +1,52 @@
+import React from 'react';
+import {View} from 'react-native';
+
+import {TextSourceSans} from '../../components/TextSourceSans';
+import { styles } from './Phase.styles';
+
+export const Phase = props => {
+  const {activePhase, futurePhases, pastPhases} = props;
+  const startDate = (futurePhases) ? futurePhases[0].start_date : null;
```
goapunk, line 9: are the parentheses needed here and in the line below?

containers/Ideas/IdeaProject.js
```diff
@@ -146,24 +139,11 @@ export const IdeaProject = (props) => {
...
-                      {activePhase.description}
-                    </TextSourceSans>
-                  </View>
-                ) : (
-                  <View style={styles.phaseContainer}>
-                    <TextSourceSans>
-                      No active phase found.
-                    </TextSourceSans>
-                  </View>
+                  <Phase
```
fuzzylogic2000, line 142: Cool! An extra phase component is a good idea!

containers/Ideas/Phase.js
```diff
@@ -0,0 +1,52 @@
...
+  const endDate = (pastPhases) ? pastPhases[pastPhases.length-1].end_date : null;
+
+  return (
+    activePhase ?  (
```
fuzzylogic2000, line 13: Yeah, all the `?` and `:` is hard to understand, but looking at it for a while it makes sense. :woman_shrugging: I am fine with leaving it like that.

containers/Ideas/Phase.js
Rineee: I thought yes, but that was wrong :sweat_smile: So removed them now, thanks!

PR #351, opened 2022-07-20
jstests/IdeaProject: testing contact infos

khamui:


containers/Ideas/IdeaProject.js
```diff
@@ -134,6 +134,7 @@ export const IdeaProject = (props) => {
                 titleStyle={visibleTab === tabs.information ? styles.tabsMenuItemActive : styles.tabsMenuItem}
                 title='Information'
                 type='clear'
+                testID='info-tab-button'
```
goapunk, line 137: I think we should avoid this :)

containers/Ideas/__tests__/IdeaProject.test.js
```diff
@@ -0,0 +1,56 @@
...
+    },
+    navigation: {
+      addListener: jest.fn(),
+      goBack: jest.fn(),
+      navigate: jest.fn()
+    }
+  };
+
+  const tree = render(<IdeaProject {...mockProject} />);
+  const infoTabBtn = tree.getByTestId('info-tab-button');
```
goapunk, line 43: using `getText('Information')` works fine for me, so I think we should remove the `testID`

containers/Ideas/__tests__/IdeaProject.test.js
```diff
@@ -0,0 +1,56 @@
...
+      },
+    },
+    navigation: {
+      addListener: jest.fn(),
+      goBack: jest.fn(),
+      navigate: jest.fn()
+    }
+  };
+
+  const tree = render(<IdeaProject {...mockProject} />);
```
goapunk, line 42: in almost all the other tests we use `const { getByText } = render(`, I'd personally prefer to use the same style everywhere :)

containers/Ideas/__tests__/IdeaProject.test.js
```diff
@@ -0,0 +1,56 @@
+import React from 'react';
+import '@testing-library/jest-native/extend-expect';
+import { render, fireEvent, act } from '@testing-library/react-native';
+import { IdeaProject } from '../IdeaProject';
+import API from '../../../BaseApi';
+
+const promise = Promise.resolve();
+jest.mock('../../../BaseApi', () => jest.fn());
```
goapunk, line 8: I'm not sure I fully understand this (but that's true for all the weird mock stuff :sweat_smile: ). One suggestion I have which seems a bit easier is to remove line 7-14 and instead just add `API.getModule = jest.fn(() => Promise.resolve());`  in the test function below. You can then just do ` await waitFor( ()=> {expect(API.getModule).toHaveBeenCalled();});` instead of the `await act(() => promise)`, what do you think?

khamui: yes i am also not following the mocking part 100%, the promise way I found from Kent C Dodds( think thats the creator of testing-lilbrary). I didn't like it actually because of the extra `promise` variable.

Codewise I like your approach more. But also technically i am not testing the getModule and therefore testing it `toHaveBeenCalled` seems really correct.

anyways, it is indeed more readable so lets go with it :) Thanks!

goapunk: hmm, good point, I didn't take that into account

goapunk (approved):
nice!

PR #363, opened 2022-07-21
WIP: [#6233] Pm 2022 07 profile with dummy data

phillimorland:
this also has design cleaning in as it was just a bit annoying


components/ButtonAvatar.js
```diff
@@ -0,0 +1,28 @@
+import React from 'react'
+import { TouchableOpacity, Image } from 'react-native'
+import { styles } from './ButtonAvatar.styles'
+
+export const AvatarCircle = (props) => {
```
goapunk, line 5: do we use that somewhere else than in this file ? If not we don't need to export it, right?

containers/Ideas/ExplorePage.js
```diff
@@ -29,9 +29,17 @@ export const ExplorePage = (props) => {
...
+    props.navigation.navigate('ProfileScreen', {name: 'Bob'})
+  }
+
   return (
     <View style={styles.container}>
-      <ButtonSignOut></ButtonSignOut>
+      <ButtonAvatar
+        labelText="profile"
+        hintText="click to go to profile and settings"
+        onPress={() => toProfile()}
```
goapunk, line 41: would  just `{toProfile}` do or do we need the arrow funtion here?

containers/User/ProfileScreen.js
```diff
@@ -0,0 +1,23 @@
+import React from 'react'
+import { View } from 'react-native'
+import { styles } from './ProfileScreen.styles'
```
fuzzylogic2000, line 3: Did you forget to commit that file?

phillimorland: dammit, yes, will push in a minute

phillimorland: @fuzzylogic2000 sorry should be good now, it's a re-write though so let me know if forgot anything

fuzzylogic2000:
I started adding the API stuff to this PR and made it WIP with that. Hope, that's fine!

phillimorland:
@fuzzylogic2000 cool, thank you! i'll go on with @goapunk review comments and update tests

fuzzylogic2000:
I made a bit of a mess with @phillimorland s PR now. And added the commit that she had to do in another branch back to this one now. 
@khamui @Rineee Could you already look at this and say what needs fixing before we merge? I think a merge would be easier to then go off into the different directions. Buuuut actually, no, the user image needs fixing for sure, when the only hard-coded user doesn't have an image this will break.

fuzzylogic2000:
Ok, I will close this in favour of #371 now and make a new branch for the API info stuff.

PR #364, opened 2022-07-22
logout if wrong token

Rineee:
I changed the permissions in backend for project and module api, so only authenticated users are allowed. I guess that makes more sense, right? It is here: https://github.com/liqd/adhocracy-plus/pull/1993

That way we can now easily check on explore page if the token is valid and log out the user, if not. I copied a bit from @goapunk s former PR :)

And I broke the frontend tests, but have no idea how to fix that :see_no_evil: Maybe someone can do that or even pair with me on it? That would be great!

### DIFF
--- a/BaseApi.js
+++ b/BaseApi.js
@@ -68,8 +68,14 @@ const makeGetRequest = (url, token = null) => {
     method: 'GET',
     headers: getHeaders(token)
   })
-    .then(response => response.json())
-    .then(unpackedData => unpackedData)
+    .then(response => {
+      const statusCode = response.status
+      const data = response.json()
+      return Promise.all([ statusCode, data ])
+    })
+    .then(values => {
+      return {statusCode: values[0], data: values[1]}
+    })
     .catch(error => console.error(error))
 }
 
@@ -156,8 +162,8 @@ const API = {
   postLogin(data) {
     return makePostRequest(endpoints.login, data)
   },
-  getProjects() {
-    return makeGetRequest(endpoints.projects)
+  getProjects(token = null) {
+    return makeGetRequest(endpoints.projects, token)
   },
   getModule(moduleId, token = null) {
     const url = endpoints.module.replace('$moduleId', moduleId)
--- a/containers/Ideas/ExplorePage.js
+++ b/containers/Ideas/ExplorePage.js
@@ -1,18 +1,30 @@
 import React, { useState, useEffect } from 'react'
 import { View, FlatList } from 'react-native'
 import API from '../../BaseApi'
+import AsyncStorage from '@react-native-async-storage/async-storage'
 import { ExploreListItem } from './ExploreListItem'
 import { styles } from './ExplorePage.styles'
 import { ButtonSignOut } from '../../components/ButtonSignOut'
 import { TextSourceSans } from '../../components/TextSourceSans'
+import {useAuthorization} from '../../containers/Auth/AuthProvider.js'
 
 export const ExplorePage = (props) => {
   const [projects, setProjects] = useState([])
+  const {signOut} = useAuthorization()
 
   const fetchProjects = () => {
-    API.getProjects()
-      .then(response => response && setProjects(response))
-      .catch(error => console.warn(error))
+    AsyncStorage.getItem('authToken')
+      .then((token) => API.getProjects(token))
+      .then((response) => {
+        if(response.statusCode === 200) {
+          setProjects(response.data)
+        } else if (response.statusCode === 401) {
+          console.warn('Unauthorized, wrong login?')
+          signOut()
+        } else {
+          return Promise.reject(new Error('fetchProjects returned ' + response.statusCode))
+        }
+      }).catch(error => console.warn(error))
   }
 
   const pressHandler = (project) =>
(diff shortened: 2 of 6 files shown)


BaseApi.js
```diff
@@ -68,8 +68,14 @@ const makeGetRequest = (url, token = null) => {
     method: 'GET',
     headers: getHeaders(token)
   })
-    .then(response => response.json())
-    .then(unpackedData => unpackedData)
+    .then(response => {
+      const statusCode = response.status
+      const data = response.json()
+      return Promise.all([ statusCode, data ])
```
khamui, line 74: does it work the same way without Promise.all() here? i think statusCode is not a Promise, but only data.

khamui: maybe more clear to return `{ statusCode, data }` and in the next `then` also return `{ statusCode, data}` instead of `values[0], values[1]`?

containers/Ideas/ExplorePage.js
```diff
@@ -1,18 +1,30 @@
...
 
 export const ExplorePage = (props) => {
   const [projects, setProjects] = useState([])
+  const {signOut} = useAuthorization()
 
   const fetchProjects = () => {
-    API.getProjects()
-      .then(response => response && setProjects(response))
-      .catch(error => console.warn(error))
+    AsyncStorage.getItem('authToken')
```
goapunk, line 16: more of a general question (to all): if all requests need auth, do we really need to write this in every file? or just move the token fetching into the API methods ?

containers/Ideas/ExplorePage.js
```diff
@@ -1,18 +1,30 @@
...
+      .then((token) => API.getProjects(token))
+      .then((response) => {
+        if(response.statusCode === 200) {
+          setProjects(response.data)
+        } else if (response.statusCode === 401) {
+          console.warn('Unauthorized, wrong login?')
+          signOut()
+        } else {
+          console.warn('fetchProjects returned ' + response.statusCode)
```
khamui, line 25: In case we want to utilize the `catch` below, from within the `then`.
we coud `Promise.reject('fetchProjects returned ' ...)`. I think this way we avoid `Unhandled Promise..` errors. But not entirely sure and not tested.

khamui (commented):
nice! just some questions and optimizations, but not a must imho

containers/Ideas/ExplorePage.js
```diff
@@ -1,18 +1,30 @@
...
+  const {signOut} = useAuthorization()
 
   const fetchProjects = () => {
-    API.getProjects()
-      .then(response => response && setProjects(response))
-      .catch(error => console.warn(error))
+    AsyncStorage.getItem('authToken')
+      .then((token) => API.getProjects(token))
+      .then((response) => {
+        if(response.statusCode === 200) {
```
goapunk, line 19: I know I did it like this but now that I look at it: should the fetch error code handling happen in BaseApi? and it either returns a results or an error? instead of doing the same error handling spread out over all the different containers?

BaseApi.js
```diff
@@ -68,8 +68,14 @@ const makeGetRequest = (url, token = null) => {
...
+    })
+    .then(values => {
+      return {statusCode: values[0], data: values[1]}
```
goapunk, line 77: the tests for `Idea.js` now fail because the mocked Api still returns a single value, not the two like here

BaseApi.js
goapunk: I copied this from the post method in BaseApi, so if we change this we should probably change it there as well.
Apart from that, as data is a Promise we do need to return a Promise, otherwise the subsequent `then` calls fail, right?

khamui: could be yes, i was just wondering if it would resolve, but probably if we have at least one promise the array we need it. sorry, you can ignore this bit @Rineee

goapunk: I think it's worth testing, as I wondered about it too! :)

Rineee: I did test that at the time I wrote it for the post method and it does not work without the promise..

BaseApi.js
Rineee: yeah, I know, I wrote in the PR description that I need help fixing the tests.. :see_no_evil:

containers/Ideas/ExplorePage.js
Rineee: I actually tried to move the checking for the 401 to BaseApi, but I could not make it work to use the signOut there (because the useAuthentication is apparently a hook (why?)  and hook usage is only allowed in classes) :see_no_evil: So yes, I think it would be better, but not sure how to do it best.. do you want to give it a go @goapunk? Or we look at it together?

containers/Ideas/ExplorePage.js
khamui: yes i think it grew overtime and now it would make sense to put it into the BaseApi

Rineee: Yeah, probably we should do that.. I think it just evolved like this.. we still need to handle the login then differently though. Maybe an issue, so we can do that in another PR?

containers/Ideas/ExplorePage.js
Rineee: Not sure I get that really.. so instead of the catch a Promise.reject()? I just thought, the catch is redundant anyway, because we already catch in BaseApi, right?

khamui: i meant if it only console.warn in `then`, then we get `Unhandled Promise...` errors.
so we should instead of `console.warn` in `then`, `Promise.reject` this way it will jump into the `catch` which we would still need. But it is also not super related to this PR, it is a more general question how we want to handle errors with fetching from the API

containers/Ideas/ExplorePage.js
goapunk: ah, right. No, I don't think we can work around that as `BaseApi` is just plain js and not React

containers/Ideas/ExplorePage.js
Rineee: @khamui I tried that out now, so I did 
`        } else {
          Promise.reject(new Error('fetchProjects returned ${response.statusCode}'))
        }
      }).catch(error => console.warn(error))
`
but that actually gives a `Unhandled promise rejection` error. Did I do that wrong maybe? 
If I leave the console.log, there is no promise error, but it just prints the warning..

khamui: i think you have to `return` it so it arrives in `.catch(error)`

Rineee: :bulb: :see_no_evil:

goapunk (approved):
cool

PR #366, opened 2022-07-25
*.styles: clean up styles to use variables and update font sizes and …

phillimorland:
…weights

removed possibility to make up font sizes and line heights as we should really avoid doing that.
added semibold so that can be correctly styled


fuzzylogic2000 (approved):
Uh, it suddenly went bold! Nice!

PR #367, opened 2022-07-25
edit ideas/adding image: disable taking picture for simulators (Apple)

khamui:


phillimorland (approved):
working nicely, just to check do we need this because it doesn't work in simulator?

khamui:
> working nicely, just to check do we need this because it doesn't work in simulator? can't test that right now, mac is too out of date, will try get it going this week for testing

exactly, its just for dev on simulators

PR #368, opened 2022-07-25
containers/ideas: disable button after idea submit

fuzzylogic2000:
I think this works, but couldn't properly test it, because the API on dev responded too fast even when I added an image.
Not sure if that would confirm it enough, but maybe we could test with adding a "wait for 10 seconds" before sending the request? What would that be in js?

### DIFF
--- a/containers/Ideas/IdeaCreate.js
+++ b/containers/Ideas/IdeaCreate.js
@@ -45,6 +45,7 @@ export const IdeaCreate = props => {
   const [initialLabels, setInitialLabels] = useState([])
   const [description, setDescription] = useState(() => editing ? idea.description : '')
   const [prevValues, setPrevValues] = useState('')
+  const [submitPending, setSubmitPending] = useState(false)
 
   useEffect(() => {
     if (error) {
@@ -105,6 +106,7 @@ export const IdeaCreate = props => {
   }
 
   const handleSubmit = (values) => {
+    setSubmitPending(true)
     values.description = description
     const formData = makeFormData(values)
     AsyncStorage.getItem('authToken')
@@ -116,32 +118,37 @@ export const IdeaCreate = props => {
         }
       })
       //error handling is provisional and should probably go somewhere else eventually
-      .then((response) => {
+      .then(response => {
         const {statusCode, data} = response
-        if (statusCode == 201) {
+        if (statusCode === 201) {
           Alert.alert('Your idea was added.', 'Thank you for participating!',  [{ text: 'Ok' }])
           props.navigation.goBack()
-        } else if (editing && statusCode == 200) {
+        } else if (editing && statusCode === 200) {
           Alert.alert('Your idea was updated.', '',  [{ text: 'Ok' }])
           props.navigation.navigate({
             name: 'IdeaDetail',
             params: { idea: data },
             merge: true,
           })
-        } else if (statusCode == 400) {
+        } else if (statusCode === 400) {
+          let errorMsg
           if (data && data.image) {
-            const errorString = data.image.reduce((acc, curr) => {
+            errorMsg = data.image.reduce((acc, curr) => {
               return `${acc}\n${curr}`
             }, 'Please try again.\n')
-            setError(errorString)
           } else {
-            setError('Required fields are missing.')
+            errorMsg = ('Required fields are missing.')
           }
-        } else if (response.statusCode == 403) {
-          setError(response.data.detail)
+          return Promise.reject(errorMsg)
+        } else if (statusCode === 403) {
+          return Promise.reject(data.detail)
         } else {
-          setError('Try again.')
+          return Promise.reject('Try again.')
         }
+      })
+      .catch(error => {
+        setError(error)
+        setSubmitPending(false)
       })
   }
 
@@ -264,7 +271,7 @@ export const IdeaCreate = props => {
               <ButtonSubmit
                 title='Submit'
                 onPress={handleSubmit}
-                disabled={!isValid}
+                disabled={!isValid || submitPending}
               >
               </ButtonSubmit>
             </>


phillimorland:
@fuzzylogic2000 you can use the `setTimeout()` function to add a delay but i'll check on the apple test device I have here and see first as thats where problem was first found

phillimorland (changes requested):
@fuzzylogic2000 still managed to press submit repeatedly before view changed to idea list when both editing and adding idea both with image. Also triggered this error trigger the error. This we might have to look at some optimisations :( 
```VirtualizedList: You have a large list that is slow to update - make sure your renderItem function renders components that follow React performance best practices like PureComponent, shouldComponentUpdate, etc. Object {
  "contentLength": 5902,
  "dt": 233353,
  "prevDt": 775,
}
```

containers/Ideas/IdeaCreate.js
```diff
@@ -110,6 +112,7 @@ export const IdeaCreate = props => {
       })
       //error handling is provisional and should probably go somewhere else eventually
       .then((response) => {
+        setSubmitPending(false)
```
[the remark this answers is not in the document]
fuzzylogic2000: But to which one? We only do not need it in the first if as we go back to the idea list then, but for all other cases we would need it to be enabled again, so that the idea can be submitted again, right?

fuzzylogic2000:
> @fuzzylogic2000 still managed to press submit repeatedly before view changed to idea list when both editing and adding idea both with image. Also triggered this error trigger the error. This we might have to look at some optimisations :(
> 
> ```
> VirtualizedList: You have a large list that is slow to update - make sure your renderItem function renders components that follow React performance best practices like PureComponent, shouldComponentUpdate, etc. Object {
>   "contentLength": 5902,
>   "dt": 233353,
>   "prevDt": 775,
> }
> ```

@phillimorland @khamui @goapunk Can someone take over? I thought, I changed the state and disabled the button right after submit and then wait for the response to enable the button again. But if that's not it, I don't know how to do it.

containers/Ideas/IdeaCreate.js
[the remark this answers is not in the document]
phillimorland: yes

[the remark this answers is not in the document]
phillimorland: it should stay disabled for the successful submit as user will be taken to different view

phillimorland:
> > @fuzzylogic2000 still managed to press submit repeatedly before view changed to idea list when both editing and adding idea both with image. Also triggered this error trigger the error. This we might have to look at some optimisations :(
> > ```
> > VirtualizedList: You have a large list that is slow to update - make sure your renderItem function renders components that follow React performance best practices like PureComponent, shouldComponentUpdate, etc. Object {
> >   "contentLength": 5902,
> >   "dt": 233353,
> >   "prevDt": 775,
> > }
> > ```
> 
> @phillimorland @khamui @goapunk Can someone take over? I thought, I changed the state and disabled the button right after submit and then wait for the response to enable the button again. But if that's not it, I don't know how to do it.

I'll take it! 
For optimisations we need to find a plan i guess,

phillimorland:
good to go I think maybe @khamui you can review it as it's a IOS issue

containers/Ideas/IdeaCreate.js
```diff
@@ -264,7 +270,7 @@ export const IdeaCreate = props => {
               <ButtonSubmit
                 title='Submit'
                 onPress={handleSubmit}
-                disabled={!isValid}
+                disabled={!isValid || submitPending && true}
```
khamui, line 273: `&& true` has no effect here, can be removed

khamui (changes requested):
It is working fine! Just one thing

containers/Ideas/IdeaCreate.js
```diff
@@ -134,13 +136,17 @@ export const IdeaCreate = props => {
               return `${acc}\n${curr}`
             }, 'Please try again.\n')
             setError(errorString)
+            setSubmitPending(false)
           } else {
             setError('Required fields are missing.')
+            setSubmitPending(false)
```
khamui, line 142: if you want to have one line less you could shift it to the end of the statement, e.g. 
```
if (data && data.image) {
  ..
} else {
  ..
}
setSubmitPending(false)
```

khamui (commented):
small optimisation, leaving to you. maybe less readable though

containers/Ideas/IdeaCreate.js
```diff
@@ -134,13 +136,17 @@ export const IdeaCreate = props => {
...
           }
         } else if (response.statusCode == 403) {
```
khamui, line 144: just saw this (it was there before), which needs to be fixed:
because we unpack it some lines above we should use `statusCode === 403` instead of `response.statusCode == 403`.

And in Javascript we want to  compare strictly which is using `===`.

khamui:
@fuzzylogic2000 @phillimorland I just pushed the fixes myself, because i found some things that needed to be fixed apart from your task. Hope that is okay. But now someone else has to check and approve. sorry.

containers/Ideas/IdeaCreate.js
```diff
@@ -118,37 +118,38 @@ export const IdeaCreate = props => {
...
+          return Promise.reject(errorMsg)
+        } else if (statusCode === 403) {
+          return Promise.reject(data.detail)
         } else {
-          setError('Try again.')
-          setSubmitPending(false)
+          return Promise.reject('Try again.')
         }
       })
+      .catch(error => {
```
goapunk, line 149: small nitpick/personal preference: move to line above instead of newline?

khamui: i think it is more consistent to have it on the same indentation as the `.then` above. Because it's chaining, it would be wrong in my opinion if we do not break it into a new line:

```
fetching()
  .then()
  .then()
  .catch()
```

vs.

```
fetching()
  .then()
    .then()
      .catch()
```

goapunk: fair!

goapunk (approved):
works nicely for me, thanks!

PR #369, opened 2022-07-25
Richtext: adding iframes as custom html and add renderer for iframes

khamui:


components/Richtext.js
```diff
@@ -29,6 +32,8 @@ export const Richtext = ({ text }) => {
   }
 
   const collapsibleRenderer = ({ TDefaultRenderer, tnode,  ...props }) => {
+    const isCollapsible = tnode.domNode.attribs.class === 'collapsible-item'
+
     if (isCollapsible(tnode)){
```
goapunk, line 37: this should now only be `isCollapsible`

goapunk, line 37: this should now be `isCollapsible` instead of a function

goapunk, line 37: this should now just be `isCollapsible` instead of a function

goapunk (changes requested):

components/Richtext.js
khamui: aiai...okay fixed

goapunk (approved):
cool

PR #371, opened 2022-07-26
Pm 2022 07 settings page

phillimorland:


fuzzylogic2000:
I cherry-picked your new commit to your old branch now. https://github.com/liqd/zt-app/pull/363

components/ButtonAvatar.js
```diff
@@ -0,0 +1,23 @@
+import React from 'react'
+import { TouchableOpacity, Image } from 'react-native'
+import { styles } from './ButtonAvatar.styles'
+
+export const ButtonAvatar = (props) => {
+
```
khamui, line 14: we could make this component responsible for if there is an image or not.
So if not, it will show a blue circle instead?

I know we want to use a default image, but it would be more robust to have a "fallback" here?

containers/User/ProfileScreen.js
```diff
@@ -0,0 +1,17 @@
...
+
+import {TextSourceSans} from '../../components/TextSourceSans'
+
+export const ProfileScreen = () => {
+  return (
+    <View>
+      <TextSourceSans>
+          profile
+      </TextSourceSans>
+      <ButtonSignOut></ButtonSignOut>
```
khamui, line 37: self-closing tag?

containers/User/SettingsOverview.js
```diff
@@ -0,0 +1,24 @@
...
+
+  const toProfileSettings = () => {
+    props.navigation.navigate('SettingsProfile', {name: 'Bob'})
+  }
+
+  return (
+    <View>
+      <Header
+        navigation={props.navigation} />
+      <TextSourceSans></TextSourceSans>
```
khamui, line 17: self-closing tag?

khamui (approved):
I added some comments, but in case we want to continue with tasks related to this PR, lets merge and move on. We can tackle this in the upcoming PR?

components/ButtonAvatar.js
fuzzylogic2000: Yes, the fallback is something, I wanted to tackle now. So, will put that into the next PR depending on this. And the other two comments, I will also add there! So, will merge now. :partying_face:

fuzzylogic2000: Actually, it already does that. When there is no image in the props (e.g. if the image is null), the blue circle is shown. Should we make that more robust and more explicit? What would be the best way? @khamui

khamui: ah okay, so we could shift that logic to this location. so in case we use the AvatarCircle somewhere else it handles that and we do not need to think about the blue circle fallback again? we can have a look together as well @fuzzylogic2000 ?

PR #372, opened 2022-07-26
CommentForm: adjusting styles to fix bug closes #265

### LINKED ISSUE #265: #5173 – Participate in idea challenge project: comment on ideas
**device & browser:** *Iphone + ExpoGo)*
**Comment/Question:**
220214-app-testing-participate-in-idea-challenge-project-comment-on-ideas

1 – The text field is touching the home swipe bar indicator. Would it be possible to make it a bit higher? I've added a WhatsApp screenshot to show how it works there.

2 – The grey text should be #666666 for accesibilty and it doesn't seem to be centred vertically. 

220214-app-testing-participate-in-idea-challenge-project-comment-on-ideas-whatsapp

khamui:
closes #265

### DIFF
--- a/containers/Comments/CommentForm.styles.js
+++ b/containers/Comments/CommentForm.styles.js
@@ -8,8 +8,7 @@ export const styles = StyleSheet.create({
     flexDirection: 'row',
     alignItems: 'center',
     paddingHorizontal: SPACINGS.multiplyBy(1),
-    paddingTop: SPACINGS.multiplyBy(0.75),
-    paddingBottom: SPACINGS.multiplyBy(2)
+    paddingVertical: SPACINGS.multiplyBy(1)
   },
   textInputContainer: {
     flexDirection: 'row',
--- a/containers/Comments/CommentForm.styles.js
+++ b/containers/Comments/CommentForm.styles.js
@@ -1,7 +1,7 @@
-import { StyleSheet } from 'react-native'
+import { Platform,StyleSheet } from 'react-native'
 
 import { COLORS } from '../../theme/colors'
-import { BORDERRADIUS,SPACINGS } from '../../theme/spacings'
+import { BORDERRADIUS, SPACINGS } from '../../theme/spacings'
 
 export const styles = StyleSheet.create({
   submitContainer: {
@@ -18,10 +18,9 @@ export const styles = StyleSheet.create({
     borderRadius: BORDERRADIUS.lg,
     padding: SPACINGS.multiplyBy(0.5),
   },
-  textInput: {
-    paddingTop: 0,
-    paddingBottom: 0
-  },
+  textInput: Platform.OS !== 'ios' 
+    ? { width: '100%' }
+    : { width: '100%', paddingTop: 0, paddingBottom: 0 },
   submitButton: {
     padding: SPACINGS.multiplyBy(0.25),
     paddingRight: SPACINGS.multiplyBy(0.4),
--- a/containers/Comments/CommentForm.styles.js
+++ b/containers/Comments/CommentForm.styles.js
@@ -8,13 +8,20 @@ export const styles = StyleSheet.create({
     flexDirection: 'row',
     alignItems: 'center',
     paddingHorizontal: SPACINGS.multiplyBy(1),
-    paddingVertical: SPACINGS.multiplyBy(0.75),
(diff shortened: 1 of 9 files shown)


containers/Comments/CommentForm.styles.js
```diff
@@ -7,14 +7,21 @@ export const styles = StyleSheet.create({
     flexDirection: 'row',
     alignItems: 'center',
     paddingHorizontal: SPACINGS.multiplyBy(1),
-    paddingVertical: SPACINGS.multiplyBy(0.75),
+    paddingTop: SPACINGS.multiplyBy(0.75),
+    paddingBottom: SPACINGS.multiplyBy(2)
```
goapunk, line 11: I guess this is more of a general thing we need to start doing: have different styles for iOS and Android because this previously looked fine on Android but not iOS. Now it's the other way round.
What we could do in most places is something like 
`paddingBottom: Platform.OS === 'ios' ? SPACINGS.multiplyBy(2) : SPACINGS.multiplyBy(0.75)`
In this particular case it's a bit more complicated I think because as I understand the screenshots in the issue we don't want the big padding if the keyboard is visible, so we have to handle that programatically ?

goapunk (changes requested):

containers/Comments/CommentForm.styles.js
goapunk: I've looked more into this general issue (I also screwed up the header on iOS) it seems we should probably use the `SafeAreaView` as described [here](https://reactnavigation.org/docs/handling-safe-area/) instead of manually guessing paddings since there could also be weird physical notches etc

khamui:
okay version with proper safeareas, tested on android emulator, iphone simulator and physical iphone (with notch)

containers/Comments/CommentForm.styles.js
```diff
@@ -7,14 +7,20 @@ export const styles = StyleSheet.create({
...
+    alignItems: 'center',
     flex: 2,
     backgroundColor: COLORS.paper,
     borderRadius: BORDERRADIUS.lg,
     padding: SPACINGS.multiplyBy(0.5),
   },
+  textInput: {
+    paddingTop: 0,
+    paddingBottom: 0
+  },
```
phillimorland, line 23: this makes the clickable area of the text input very small clickable area should be at minimum 40x40

phillimorland (commented):
very nice! if we can make component as discussed would be great but if not then this is good to merge other then clickable area

containers/Comments/CommentForm.styles.js
khamui: for me it does not change the clicable area, if i do not set it to 0 it is not vertically centered.

goapunk: the padding doesn't do anything for me (looks the same with and without). however, the clickable area is only as wide as the placeholder text, clicking right of the placeholder text does nothing. Adding something like `width: '100%'` fixes it for me

khamui: seems to be an iOS issue only, latest commit fixed it.

goapunk:
@khamui this unfortunately has a lot of merge conflicts now :/

containers/Ideas/IdeaCreate.js
```diff
@@ -16,6 +16,11 @@ import {
 import { LabelList,LabelListContainer } from '../../components/LabelForm'
 import { TextSourceSans } from '../../components/TextSourceSans'
 import { VirtualScrollView } from '../../components/VirtualScrollView'
+<<<<<<< HEAD
```
goapunk, line 19: rebase went wrong here ;-)

containers/Ideas/IdeaProject.js
```diff
@@ -5,7 +5,15 @@ import IconSLI from 'react-native-vector-icons/SimpleLineIcons'
 import AsyncStorage from '@react-native-async-storage/async-storage'
 import { Button } from '@rneui/themed'
 import { LinearGradient } from 'expo-linear-gradient'
+<<<<<<< HEAD
```
goapunk, line 8: rebase went wrong here ;-)

khamui: uff, sorry! and thanks for catching it..

goapunk (approved):
looks  good to me, I fixed the failing tests

PR #374, opened 2022-07-26
ExplorePage and ProfilePage: add user info

fuzzylogic2000:
depends on https://github.com/liqd/adhocracy-plus/pull/1995

I will add the user fallback stuff (if at all) in a different PR.


Rineee:
> @Rineee I already added your API. Maybe you can also go on from here? Even though that adds more PRs that have to be finished and merged before that one..

OK! So this is ready to be reviewed?

fuzzylogic2000:
> > @Rineee I already added your API. Maybe you can also go on from here? Even though that adds more PRs that have to be finished and merged before that one..
> 
> OK! So this is ready to be reviewed?

@Rineee Not really, I think, it's just the start and needs fixing in the first three commits in the other PR first. But if you have comments for the last commit, please already add that!

Rineee (commented):
Really cool, working nicely! 

But, the user image is half behind the whiter header for me.. it is on Android Pixel 4 phone. Not sure, if that was already there before though?

And I am bit confused with the screen in the middle between the profile and settings. But I guess that is because it is almost empty..

Rineee:
> But, the user image is half behind the whiter header for me.. it is on Android Pixel 4 phone. Not sure, if that was already there before though?
> 
Yeah, so that was there before, so it is rather an issue, I guess? I will add one.

So good to go from my side, someone else wants to have a look @khamui @goapunk ?

fuzzylogic2000:
> Really cool, working nicely!
> 
> But, the user image is half behind the whiter header for me.. it is on Android Pixel 4 phone. Not sure, if that was already there before though?

Yeah, that was probably merged like that with the last PR. But we still need to fix that. :p
> 
> And I am bit confused with the screen in the middle between the profile and settings. But I guess that is because it is almost empty..

Eh, what screen in the middle? There should be:
1. the new user image of the authenticated user in the upper right corner of the explore page
2. the profile page (that looks quite empty as long as we only have image and username), with the settings-thingy in the upper right corner
3. the actual settings page where you should then be able to chose different things to change like account stuff (planned for this sprint) and later also password or organisation terms of use agreements

Actually, writing that down, I need to add, that 

- [ ] the profile screen should also only have the logout button when the user is_self!

Rineee:
> Eh, what screen in the middle? There should be:
> 
>     1. the new user image of the authenticated user in the upper right corner of the explore page
> 
>     2. the profile page (that looks quite empty as long as we only have image and username), with the settings-thingy in the upper right corner
> 
>     3. the actual settings page where you should then be able to chose different things to change like account stuff (planned for this sprint) and later also password or organisation terms of use agreements
> 
> 
> Actually, writing that down, I need to add, that
> 
>     * [ ]  the profile screen should also only have the logout button when the user is_self!
> 
> 
> @Rineee

Yes, I meant the settings page, that confused me, but once we add more stuff there, it will make sense! So all good!

containers/Ideas/ExplorePage.js
```diff
@@ -27,6 +28,18 @@ export const ExplorePage = (props) => {
...
+  const fetchAuthenticatedUser = () => {
+    AsyncStorage.getItem('authToken')
+      .then((token) => API.getAuthenticatedUser(token))
+      .then((response) => {
+        if(response.statusCode === 200) {
+          setUser(response.data)
+        } else {
+          return Promise.reject(new Error('fetchAuthenticatedUser returned ' + response.statusCode))
+        }
+      }).catch(error => console.warn(error))
```
khamui, line 40: little nit-picky here, in another PR we decided to put `.catch` on the new line because of:

copying the entire comment here (because linking lead to something else):

> i think it is more consistent to have it on the same indentation as the `.then` above
> 
> ```
> fetching()
>   .then()
>   .then()
>   .catch()
> ```
> 
> vs.
> 
> ```
> fetching()
>   .then()
>   .then().catch()
> ```

containers/Ideas/ExplorePage.js
```diff
@@ -41,17 +54,24 @@ export const ExplorePage = (props) => {
...
   return (
     <View style={styles.container}>
-      <ButtonAvatar
-        labelText="profile"
-        hintText="click to go to profile and settings"
-        onPress={toProfile}
-      ></ButtonAvatar>
+      {user &&
+        <ButtonAvatar
+          imgSource={{ uri: user._avatar }}
```
khamui, line 69: so that works for local and dev now?
at least in iOS it shows the avatar connected to local api.

khamui (commented):
Really cool! works great, just some small stuff

containers/Ideas/ExplorePage.js
fuzzylogic2000: Yes, the user API currently gives the whole URL, so it does work. 
I only checked the comments to find out about the fallback and was confused, because that API does not give the whole URL, but only the bit after the base. I will dive into that deeper for the fallback, though.

khamui:
@fuzzylogic2000 did you push something new?

fuzzylogic2000:
> @fuzzylogic2000 is the latest code missing?

@khamui Yes, I added that to another branch. Have to figure out how to get it into this one! :see_no_evil:

fuzzylogic2000:
@khamui Now it's there! Could you have another look?

containers/User/ProfileScreen.js
```diff
@@ -21,20 +23,43 @@ export const ProfileScreen = (props) => {
...
+  const fetchUser = () => {
+    return AsyncStorage.getItem('authToken')
+      .then((token) => API.getUser(userId, token))
+      .then((response) => {
+        if(response.statusCode === 200) {
+          setUser(response.data)
+        } else {
+          return Promise.reject(new Error('fetchUser returned ' + response.statusCode))
+        }
+      }).catch(error => console.warn(error))
```
khamui, line 38: the `.catch` here too 😬

khamui (commented):
@fuzzylogic2000

containers/User/ProfileScreen.js
fuzzylogic2000: Good catch! :)

Issue #350, opened 2022-07-20
back button on ios iphone SE


phillimorland:
back button has too little top padding on ios device so half overlaps with status stuff

PR #375, opened 2022-07-26
Jd 2022 07 auto focus on edit + enable swipe gestures on iOS

### LINKED ISSUE #354: #6161 UX/UI testing - edit comment
**device & browser:** *iOS 15.5*
**Comment:** *I’ll add more than one thing here + screenshots, I hope it is fine. 

01 - When editing or writing a comment, it’s important to take in consideration the bottom spacing on iOS otherwise it stays at the same spot as the bar.
IMG_4694

02 - The send blue button seems to be a bit cropped in a way that it doesn’t look like a full elipse. There is a svg on Zeplin which is a full elipse.*
IMG_4694

03 - Would it be possible to make it in a way that when the user click Edit, the comment already opens with the keyboard like ready to type, instead of having to click on the comment again?* 
IMG_4695

goapunk:
partly fixes #354 and #350

### DIFF
--- a/containers/Ideas/Idea.js
+++ b/containers/Ideas/Idea.js
@@ -1,5 +1,6 @@
 import React, { useState, useEffect, useRef } from 'react'
-import { Alert, View, Image, ScrollView, Platform, KeyboardAvoidingView, Pressable } from 'react-native'
+import { Alert, View, Image, ScrollView,
+  Platform, KeyboardAvoidingView, Pressable } from 'react-native'
 import { Button } from '@rneui/base'
 import { styles } from './Idea.styles'
 import IconSLI from 'react-native-vector-icons/SimpleLineIcons'
@@ -23,7 +24,9 @@ export const Idea = (props) => {
   const [deleteModalVisible, setDeleteModalVisible] = useState(false)
   const [processing, setProcessing] = useState(false)
   const [comments, setComments] = useState([])
-  const [contentObjectOfComment, setContentObjectOfComment] = useState({'contentType': idea.content_type, 'pk': idea.pk})
+  const [contentObjectOfComment, setContentObjectOfComment] = useState(
+    {'contentType': idea.content_type, 'pk': idea.pk}
+  )
   const [commentLastCommented, setCommentLastCommented] = useState(-1)
   const [isScrolling, setIsScrolling] = useState(false)
   const [isEditing, setIsEditing] = useState(false)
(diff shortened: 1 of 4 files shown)


containers/Ideas/Idea.js
```diff
@@ -291,6 +295,15 @@ export const Idea = (props) => {
     fetchComments(content_type, pk)
   }, [idea])
 
+  useEffect(() => {
+    if(isEditing) {
+      const timer = setTimeout(()=>{
```
goapunk, line 300: the timeout is somehow necessary, without the small delay it doesn't work

containers/Ideas/Idea.js
```diff
@@ -140,6 +143,7 @@ export const Idea = (props) => {
 
   const toggleEditing= (comment) => {
     if (isEditing){
+      commentInputRef.current.blur()
```
goapunk, line 146: hide keyboard when pressing back

containers/Ideas/Idea.js
```diff
@@ -312,6 +326,7 @@ export const Idea = (props) => {
         <ScrollView
           style={styles.container}
           scrollEnabled={!isEditing}
+          keyboardShouldPersistTaps='handled'
```
goapunk, line 329: needed because otherwise the scrollview will swallow the tap if the keyboard is open

containers/Ideas/Idea.js
```diff
@@ -304,6 +317,7 @@ export const Idea = (props) => {
       <Header
         isEditing={isEditing}
         rightButton={rightHeaderButton}
+        toggleEditing={toggleEditing}
         navigation={props.navigation} />
```
khamui, line 321: should we break this closing bracket on the next line? i mean generally?

```
<Compname
   attrib1
   attrib2
   ...
   attribN
/>
```

khamui (approved):
Very cool. The ios swipe is nice!
Just some comments about code formatting. Not related to this here, but i think we should start making lines shorter than 100 chars in those files which we are working on.

fuzzylogic2000 (approved):
Uh, that's nice! (I could not test the swipe, though.)

containers/Ideas/Idea.js
```diff
@@ -1,5 +1,6 @@
 import React, { useState, useEffect, useRef } from 'react'
-import { Alert, View, Image, ScrollView, Platform, KeyboardAvoidingView, Pressable } from 'react-native'
+import { Alert, View, Image, ScrollView,
```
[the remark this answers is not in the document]
goapunk: we should have an eslint config for that. I can open a PR for that

goapunk:
@khamui @fuzzylogic2000 should we merge this ?

PR #376, opened 2022-07-26
WIP using keyboardavoidingview to manage keyboard for android and ios

khamui:


khamui:
- [ ] Android: initial opening screen
- [ ] iOS: initial opening screen
- [ ] Android: clicking outside to close keyboard
- [ ] iOS: clicking outside to close keyboard
- [ ] Android: clicking inside to open keyboard again
- [ ] iOS: clicking inside to open keyboard again

khamui:
Note: in Apple Simulator sometimes using Virtual Keyboard is deactivated. To activate press shortcut `shift` + `cmd` + `k` in the simulator

phillimorland (commented):
Very nice! 
Just a question, do we want to add it as a component? we also have this issue on the login screen and add idea screen but not sure if we want it there? however we will want it for the username and password updating.

phillimorland:
also getting these warning but they seem to pop up every now and again
```
You specified `onScroll` on a <ScrollView> but not `scrollEventThrottle`. You will only receive one event. Using `16` you get all the events but be aware that it may cause frame drops, use a bigger number if you don't need as much precision.
```
```
You specified `onScroll` on a <ScrollView> but not `scrollEventThrottle`. You will only receive one event. Using `16` you get all the events but be aware that it may cause frame drops, use a bigger number if you don't need as much precision.
```

containers/Reports/ReportCreateMessage.styles.js
```diff
@@ -14,4 +14,10 @@ export const styles = StyleSheet.create({
     fontWeight: FONTWEIGHT.bold,
     paddingHorizontal: SPACINGS.multiplyBy(0.75),
   },
+  keyboardContainer: {
+    flex: 1
+  },
+  scrollContainer: {
+    flex: 1
```
phillimorland, line 21: you can just make and use `flexContainer` for both here

containers/Ideas/IdeaCreateDescription.js
```diff
@@ -68,7 +80,7 @@ export const IdeaCreateDescription = props => {
               disabled={!isValid}
             >
             </ButtonSubmit>
```
fuzzylogic2000, line 83: Even though not related, but while you are near that in the code, this could be self-closing. :p

fuzzylogic2000 (approved):
Nice! But yes, let's make it a component? I think, we also need it for the login...

khamui:
sorry, i am not following @phillimorland & @fuzzylogic2000 . What should I turn into a component?

phillimorland:
@khamui have a container component to be put everything in that can be used whenever there is an input field, this way we are guaranteed same behaviour everywhere, doe that make sense?

containers/Ideas/IdeaCreateDescription.js
```diff
@@ -60,6 +71,7 @@ export const IdeaCreateDescription = props => {
                 touched={touched.description}
                 numberOfLines={10}
                 multiline={true}
+                autoFocus
               />
             </ScrollView>
```
khamui, line 77: @phillimorland just figured out that I cannot put KeyboardAvoidingView and ScrollView into one component, because between those wrappers, we have this ButtonSubmit in this case

phillimorland: @khamui as we will always use this for inputs can we not add that to the component? the point is to get the same behaviour on all text submit screens to it kind of makes sense unless i'm missing something?

khamui: uff yeah make sense, sorry my brain is lagging today

phillimorland (approved):
@khamui super cool!!! and working like a charm, it can also be added to the IdeaCreate and the Login screens, once thats done it's good to be merged

components/KeyboardScrollView.js
```diff
@@ -0,0 +1,46 @@
...
+import { KeyboardAvoidingView, Platform,ScrollView } from 'react-native'
+
+import { ButtonSubmit } from './ButtonSubmit'
+import { styles } from './KeyboardScrollView.styles'
+
+export const KeyboardScrollView = (props) => {
+  const stylesOverride = props.style
+    ? {...styles, ...props.style}
+    : styles
+
```
phillimorland, line 11: is there an advantage to writting it like this over `[styles, props.style]` like we do elsewhere? it's the same right or am I being dumb?

phillimorland (approved):
cool, one questions but will merge anyway also updated the snapshots and couple of styling oddities but they should be fixed in @goapunk PR

PR #377, opened 2022-07-28
#6234 List: create and style component for settings list and links in overv…

phillimorland:
…iew and add to view

SettingsProfile: add styling to profile page

NOTE: image save is to be done in separate PR by Rine


BaseApi.js
```diff
@@ -217,7 +218,11 @@ const API = {
   getAuthenticatedUser(token = null) {
     const url = endpoints.account
     return makeGetRequest(url, token)
-  }
+  },
+  editUser(userPk, data, token = null) {
+    const url = endpoints.accounts.replace('$userPk', userPk)
+    return makePutRequest(url, data, token)
+  },
```
phillimorland, line 225: getting a 400 would like to pair to understand why a bit more

Rineee: The account url doesnt need a user pk, it always operates on the logged in user. I made it that way to get the avatar on the ExplorePage, where we dont have the user pk yet..

BaseApi.js
```diff
@@ -51,7 +51,8 @@ const endpoints = {
   ratings: baseApiUrl + '/contenttypes/$contentTypeId/objects/$objectPk/ratings/',
   rating: baseApiUrl + '/contenttypes/$contentTypeId/objects/$objectPk/ratings/$ratingId/',
   report: baseApiUrl + '/reports/',
-  user: baseApiUrl + '/users/$userPk/'
+  user: baseApiUrl + '/users/$userPk/',
+  accounts: baseApiUrl + '/account/$userPk/',
```
Rineee, line 55: So here it would only be `baseApiUrl + ''/account/'`

Rineee: ah, and I just saw that the url is already there above!

containers/User/SettingsProfile.js
```diff
@@ -1,24 +1,120 @@
...
+      .then((response) => {
+        if(response.statusCode === 200) {
+          setUser(response.data)
+        } else {
+          return Promise.reject(new Error('fetchUser returned ' + response.statusCode))
+        }
+      })
+      .catch(error => console.warn(error))
+  }
+
```
phillimorland, line 62: @Rineee as I need to update the input placeholder I re-fetched the user here but is there a more efficient way to do it? i didn't think so but wanted to check

Rineee: I think the update should also give you back the user, so you could just use the response.date from edit response and setUser to that data, would that work?

Rineee: Ah, I think I got it wrong, you maybe meant initially when you get on settings page? Cant we pass the the data there?

phillimorland: yep, I was passing the data but then `ProfileSettings` data wasn't updating after save as it was still using the data passed from `Profile`
But you're saying to pass data for initial render and get updated data from response, however i'm not sure how to tell it which to use, as for initial rendering it will use the data from 'Profile' again, does that make sense?

goapunk:
@phillimorland I rebased and fixed the issues with the header which came with it. Do you want to review them? If not you can just merge

goapunk (approved):
nice!

Issue #378, opened 2022-07-28
Refactor: improve error handling for all fetch requests


khamui:
Expected: after fetching successes/errors are handled centralized, not in every component again
Current Behavior: every fetching component is handling responses individually

hello-kathi:
@khamui is this issue relevant when we want to have the app test ready? Then I would put into the milestone.

khamui:
@hello-kathi It is more of an improvement/refactor. I would say not relevant as it does not add functionality for the user.

PR #380, opened 2022-07-28
#6256 button disable fetch: comments & reporting buttons (edit idea works already)

khamui:


containers/Ideas/Idea.js
```diff
@@ -1,5 +1,13 @@
 import React, { useState, useEffect, useRef } from 'react'
-import { Alert, View, Image, ScrollView, Platform, KeyboardAvoidingView, Pressable } from 'react-native'
+import {
+  Alert,
```
fuzzylogic2000, line 3: While you are at that: can we also order the imports alphabetically?
And maybe for the imports also have a rule like 
1. React
2. react-native
3. other stuff
4. own stuff - components
5. own stuff - styles
(Just as an idea for a rule, but I find the imports a bit hard to read if we have many.)

containers/Ideas/Idea.js
```diff
@@ -54,10 +75,13 @@ export const Idea = (props) => {
...
           'ReportCreateMessage',
-          {content_type: ideaState.content_type, object_pk: ideaState.pk}
+          {
+            content_type: ideaState.content_type,
+            object_pk: ideaState.pk
+          }
         )
       },
-      isFirst: !ideaState.has_changing_permission && !ideaState.has_deleting_permission,
+      isFirst: noPermissions,
```
fuzzylogic2000, line 86: if we use the noPermissions variable only once, I like the old version more as it is easier to read - with the rules directly where they are used.

fuzzylogic2000 (approved):
Nice!

phillimorland (commented):
looking good, just couple comments

containers/Ideas/Idea.js
khamui: do you mean not to use `noPermissions`? the reason is what @fuzzylogic2000 wrote, that it is easier to read inline and it is used only once

containers/Ideas/Idea.js
```diff
@@ -180,94 +181,65 @@ export const Idea = (props) => {
...
-          contentObjectOfComment.contentType,
-          contentObjectOfComment.pk,
-          values,
-          token
-        )
+        const payload = getCommentPayload(values, token)
+
+        return isEditing
+          ? API.editComment(...payload)
+          : API.addComment(...payload)
```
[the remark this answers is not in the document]
khamui: this was a bigger refactor actually, we used two functions that did almost the same and it was quite hard to add or change things, because you had to do it for both. I think now it is better separated and easier to maintain. But sure, i can revert it if the majority prefers the other way.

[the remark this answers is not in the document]
phillimorland: it's more about the payload, I just find it easier to see what's happening when it's in the submit but also happy to go with majority

containers/Ideas/Idea.js
phillimorland: sorry, I thought it still read the `noPermission` thing and was unresolved

containers/Ideas/Idea.js
```diff
@@ -313,6 +331,12 @@ export const Idea = (props) => {
     }
   }, [isEditing])
 
+  useEffect(() => {
```
goapunk, line 334: What's the advantage/reason of putting the error handling in a hook instead of doing it where the error happens?

containers/Ideas/Idea.js
```diff
@@ -162,88 +181,83 @@ export const Idea = (props) => {
...
+    setContentObjectOfComment({
+      'contentType': commentContentType,
+      'pk': commentObjectPk
+    })
     commentInputRef.current.focus()
   }
 
   function handleScroll(event) {
-    if (isScrolling && event.nativeEvent.contentOffset.y < 75.0) {
+    const { nativeEvent: { contentOffset: { y: offsetY } } } = event
```
goapunk, line 257: I find this somehow harder to read than just `const offsetY = event.nativeEvent.contentOffset.y. Is there any advantage to this style?

containers/Ideas/Idea.js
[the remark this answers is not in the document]
khamui: @phillimorland do you have a concrete suggestion how i could make it more readable here?

containers/Ideas/Idea.js
khamui: i think just the matter of taste. Maybe because i am used to this notation. Will change it. Important to me is to "unpack" the variable `y` instead of having it in the condition.

containers/Ideas/Idea.js
khamui: the reason is, that we use the hook way in IdeaCreate:
https://github.com/liqd/zt-app/blob/bfbac203a16e65da3c7b55f41267fdf6ecd8683e/containers/Ideas/IdeaCreate.js#L51

and the advantage is that it "centralizes" the `Alert.alert()` call. If there will be more error throws in future we can just set the `error` state variable. Imho makes it more consistent and easier to extend. Does that make sense to you?

goapunk: ok, didn't know we do it that way elsewhere too. I was just a bit concerned with state being async and everything that it makes it a bit more complicated, it could also be centralized in a function as long as we don't pass it down to subcomponents. I'm also fine with the hook though!

containers/Ideas/Idea.js
goapunk: Fair, you can also leave it like this if you prefer it

containers/Ideas/Idea.js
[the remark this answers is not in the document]
goapunk: It's ok like this for me, I personally would prefer a simple `if` as I'm more used to that:
```
if (isEditing) return API.editComment(...payload)
return API.addComment(...payload)

goapunk (approved):
Cool, looks good to me. Will merge once you resolved the merge conflicts :)

containers/Ideas/Idea.js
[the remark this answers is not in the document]
phillimorland: I agree that we should use `if` over ternary which are less readable, I thought we decided to only use ternary in simple cases, like where everything can be on 1 line?!

containers/Ideas/Idea.js
khamui: yes that is true. I just saw that we also use the hook way for the `LoginScreen`. If you don't mind, I would leave it for now like that and revisit the error handling, when we work on this issue?

https://github.com/liqd/zt-app/issues/378

containers/Ideas/Idea.js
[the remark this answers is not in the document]
khamui: yes true, okay i will get rid of the ternary notation also in `getCommentPayload`

containers/Ideas/Idea.js
goapunk: sounds good to me, thanks :)

goapunk:
cool, thanks! and sorry for the back and forth

containers/Ideas/Idea.js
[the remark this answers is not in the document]
Rineee: Sorry for this late comment, but tbh I find the above solution more readable and nicer structured.. :see_no_evil: Totally personal preferences, maybe before next app sprint, we can talk about that again.

Issue #263, opened 2022-02-14
[Android & iOS] IdeaEdit: Adding imaging later is not possible


khamui:
**URL:** 
**user:** 
**expected behaviour:** Editing an idea in order to add an image should be able
**behaviour:** Editing an idea and adding an image after throws an error
**important screensize:**
**device & browser:** 
**Comment/Question:** 

Screenshot?
<img width="404" alt="Bildschirmfoto 2022-02-14 um 10 47 45" src="https://user-images.githubusercontent.com/52801593/153856590-397b8b3d-ed34-4017-be45-6406b7d4277b.png">

Rineee:
Note: works for me on both iOS and Android.. lets check in next app sprint @khamui!

Rineee:
https://sentry.liqd.net/organizations/liqd/issues/2383/?alert_rule_id=30&alert_timestamp=1658332991459&alert_type=email&environment=production&project=30&referrer=alert_email

khamui:
I just checked on the iPhone (Simulator) and figured out two things.

1.) Taking a photo on the iphone simulator is not possible, so I will add a commit to disable it for Apple Simulators
2.) The following screenshot shows an error that happens if picking images from the phone's library. But this error only happens for some of the "wallpaper" photos. I receive a 500 and i am not sure what is the problem there.

![Bildschirmfoto 2022-07-21 um 11 03 31](https://user-images.githubusercontent.com/52801593/180175662-deb0464c-4859-422d-ac18-cb27aa164ef1.png)

@Rineee maybe we can pair on this?

khamui:
exact same issue happens also if using a real iphone device and taking a foto ⬆️

PR #382, opened 2022-07-29
Ideas/IdeaCreate: fix image sending errors

Rineee:
needs https://github.com/liqd/adhocracy-plus/pull/2003

This hopefully fixes https://github.com/liqd/zt-app/issues/263 as it handles image deletion differently.

There was also a bug, that if an idea with an image was edited without editing the image, the image was deleted on submit :see_no_evil:

### DIFF
--- a/containers/Ideas/IdeaCreate.js
+++ b/containers/Ideas/IdeaCreate.js
@@ -32,7 +32,7 @@ export const IdeaCreate = props => {
     imageChecked: yup
       .boolean()
       .when('image', {
-        is: (image) => image !== null,
+        is: (image) => !!image,
         then: yup.boolean()
           .oneOf([true], 'Please confirm the copyright')
       })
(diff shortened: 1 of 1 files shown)


containers/Ideas/IdeaCreate.js
```diff
@@ -266,7 +275,7 @@ export const IdeaCreate = props => {
                 }}
                 onIconPress={() => setFieldValue('imageChecked', !values.imageChecked)}
                 checked={values.imageChecked}
-                image={values.image ? values.image.uri : (idea && idea.image)}
+                image={(values.image && values.image.uri) ? values.image.uri : (values.image && values.image)}
```
Rineee, line 278: This is needed because, if a new image was uploaded, it is an object with uri property, but if no new image has been uploaded, the value is set to idea.image (happens in getInitialValues) and that is already the uri of the image.

khamui:
that works for Apple simulators now, at least i receive an alert saying that it exceeds the max. allowed file size.
but in case I want to delete an image, it shows the following error:

Bildschirmfoto 2022-08-02 um 12 26 52

@Rineee do you know what that means?

phillimorland (approved):
working perfectly on test device for both editing and deleting images so will merge and we can see how it goes

PR #384, opened 2022-08-02
Jd 2022 08 fix header status bar

### LINKED ISSUE #350: back button on ios iphone SE

goapunk:
closes #350

### DIFF
--- a/containers/Ideas/IdeaCreate.styles.js
+++ b/containers/Ideas/IdeaCreate.styles.js
@@ -7,7 +7,9 @@ import { SPACINGS } from '../../theme/spacings'
 export const styles = StyleSheet.create({
   container: {
     paddingHorizontal: SPACINGS.multiplyBy(0.75),
-    backgroundColor: COLORS.paper
+  },
+  header: {
+    paddingTop: SPACINGS.multiplyBy(0.25)
   },
   title: {
     fontSize: SIZES.md,
@@ -21,6 +23,7 @@ export const styles = StyleSheet.create({
     color: COLORS.text,
   },
   flexContainer: {
-    flex: 1
+    flex: 1,
+    backgroundColor: COLORS.paper
   }
 })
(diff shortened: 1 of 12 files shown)


containers/Ideas/IdeaProject.js
```diff
@@ -87,11 +87,11 @@ export const IdeaProject = (props) => {
         source={{ uri: bgImage }}
         accessibilityIgnoresInvertColors={true}
       />
+      <LinearGradient
```
goapunk, line 90: I'm not super sure about this but it really looks weird if we don't move it here. Or maybe we need to have a negative margin on the scrollview or something

containers/Ideas/IdeaCreate.styles.js
```diff
@@ -7,7 +7,9 @@ import { SPACINGS } from '../../theme/spacings'
 export const styles = StyleSheet.create({
   container: {
     paddingHorizontal: SPACINGS.multiplyBy(0.75),
-    backgroundColor: COLORS.paper
+  },
+  header: {
+    paddingTop: SPACINGS.multiplyBy(0.25)
```
phillimorland, line 12: I wonder if this should just be on the header and we unify the top padding on all containers so we don't have it set individually all over the place?

containers/Reports/ReportCreateMessage.js
```diff
@@ -46,8 +47,10 @@ export const ReportCreateMessage = props => {
   }
 
   return (
-    <View style={styles.container}>
-      <Header navigation={props.navigation} />
+    <SafeAreaView style={styles.container}>
+      <View style={styles.header}>
```
phillimorland, line 51: same as above about unifying padding

phillimorland (approved):
works like a charm, just question about trying to unify padding but I don't feel it's important enough not to merge but we should look into it

PR #391, opened 2022-08-04
Update currentState and docs

phillimorland:


README.md
```diff
@@ -14,11 +14,14 @@ Setup
 - Clone repo
 - Change directory into project folder `cd zt-app`
 - `make install`
-- If you do `make start` now, a window will open in your browser. If you click "Run on Andoird" or "Run on iOS", it will tell you how set these up.
+- If you do `make start` now, a window will open in your browser. If you click
+"Run on Andoird" or "Run on iOS", it will tell you how set these up. You can use
```
goapunk, line 18: not your type but: Andoird -> Android

goapunk (approved):
nice, thanks!

README.md
```diff
@@ -14,11 +14,14 @@ Setup
...
+`make start-web` if you want to view in chrome however some styling will be broken.
+
 
 Setup after Hibernation
-- Check correct Node version: 18.3.0 (15.6.22) -> Note: expo-cli has not yet been tested against v18.3.0
-- Update expo-cli - `npm install -g expo-cli`: 5.4.9 (15.6.22)
+- Check correct Node version: 18.4.0 (04.08.22) -> Note: expo-cli has not yet been tested against v18.4.0
```
khamui, line 23: was wondering if it is better to include the `Technical Review (Updates)` part into this one?

docs/coding_guidelines.md
```diff
@@ -41,12 +41,15 @@ import { MyComponent } from './path';
...
+to make bigger areas such as tiles and images and text areas pressable, in this
+case additional attributes must be added to ensure a11y, `accessibilityRole="button"` ect.
+we have implemented `eslint-plugin-react-native-a11y` to assist with this.
+The `Buttons` we use are react-elements buttons in most cases, so a third party
+button element, which has the a11y built in.
 
 ## // design patterns and rules for refactoring
 **Stateless/Stateful**
-When should we use functional components and when should we use class components? Technically it is fine to use both interchangeably in React. But somehow it feels unnecessarily random.
+Following the evolution of React, only functional components should be used.
```
khamui, line 52: Maybe just this is enough now?

```
**Component notation**
Following the evolution of React, only functional components should be used.
```

And remove the following sections?

```

So wouldn't it be nice to have a pattern which we follow (most of the time)? In this case the proposal is separating components into two different types.

Functional components (stateless)
Container components (stateful)
Functional components:

“Dumb components are also called ‘presentational’ components because their only responsibility is to present something to the DOM. Once that is done, the component is done with it. No keeping tabs on it, no checking in once in a while to see how things are going. Nope. Put the info on the page and move on.”

Container components:

“Smart components (or container components) on the other hand have a different responsibility. Because they have the burden of being smart, they are the ones that keep track of state and care about how the app works. Using the container design pattern, the container components are separated from the presentational components and each handles their own side of things. The container components do the heavy lifting and pass the data down to the presentational components as props.”

Source of excerpts: https://medium.com/@thejasonfile/dumb-components-and-smart-components-e7b33a698d43

```

khamui (commented):
Really nice! thanks

README.md
phillimorland: I'll move it right above but won't squash it into one as we all need to do the setup but only one of us does the technocal review

khamui:
nice!

PR #392, opened 2022-08-04
Issues: KeyboardScroll: adding component to all relavent pages with same stru…

phillimorland:
…cture to ensure consistent behaviour

final issue fixes to ensure idea adding is useable before putting it to bed

### DIFF
--- a/components/KeyboardScrollView.js
+++ b/components/KeyboardScrollView.js
@@ -5,9 +5,6 @@ import { ButtonSubmit } from './ButtonSubmit'
 import { styles } from './KeyboardScrollView.styles'
 
 export const KeyboardScrollView = (props) => {
-  const stylesOverride = props.style
-    ? {...styles, ...props.style}
-    : styles
 
   const buttonText = props.buttonText
     ? props.buttonText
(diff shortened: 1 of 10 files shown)


components/KeyboardScrollView.js
```diff
@@ -29,10 +26,10 @@ export const KeyboardScrollView = (props) => {
     <KeyboardAvoidingView
       behavior={behavior}
       keyboardVerticalOffset={defaultVerticalOffset}
-      style={stylesOverride}
```
phillimorland, line 32: As we always need this to be the same as it ensure button is at bottom of screen I just removed it so we aren't repeating lots of code

Rineee (changes requested):
Nice refactor!

But now it doesnt pick up on the initial values, i.e. labels and categories are not set (before on create, there was always the first category pre selected, so we dont need to validate that the category is actually set) and when I enter a title and then go to description screen and back, it tells me I need to set a title.

phillimorland:
@Rineee so this has kind of fixed it but styling for the submit button always being at bottom of page is broken but as we 'need' this structure for the dropdown for now, I think this is best solution without going down a rabbit hole

PR #393, opened 2022-08-04
Ks 2022 08 comments and profile pic

Rineee:
Depends on both

https://github.com/liqd/adhocracy4/pull/1176
https://github.com/liqd/adhocracy-plus/pull/2018

and they need to be merged first in that order!

To test, you only need to install the aplus branch as that has the a4 update hash already in it.

Quesions/things to do, but I would say a thing for later? So maybe we can put that in the hibernation docs.. 

- [ ] show fallback image in settings form (not sure if really necessary?)
- [ ] do we want to enable deletion of profile pic

### DIFF
--- a/containers/Comments/Comment.js
+++ b/containers/Comments/Comment.js
@@ -1,11 +1,11 @@
 import React, { useCallback, useEffect, useRef,useState } from 'react'
-import { Alert, Image,View } from 'react-native'
+import { Alert, Image, View } from 'react-native'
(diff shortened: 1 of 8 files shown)


khamui (approved):
it is working fine!

Just some comments, but more of code formatting. Can be merged in case we want to postpone those changes.

khamui:
merging and leaving comments for later!

PR #397, opened 2022-08-08
jest tests

phillimorland:


containers/Reports/__tests__/ReportCreateMessage.test.js
```diff
@@ -0,0 +1,73 @@
...
+})
+
+// button not disabled
+// test('Test ReportCreateMessage, no charactures button disabled', async () => {
+//   const route = { params: {
+//     content_type: testIdeaAdmin.content_type,
+//     object_pk: testIdeaAdmin.pk
+//   }}
+//   const description = ''
+//   render(<ReportCreateMessage route={route} description={description}/>)
```
phillimorland, line 55: Do i need to add the `isValid` logic here to ensure button is disabled? I thought adding condition that would disable it should be enough but also feel I haven't quite got the testing stuff

containers/User/__tests__/SettingsProfile.test.js
```diff
@@ -0,0 +1,58 @@
...
+  const nameInput = screen.getByLabelText('Username')
+
+  await act(async () => {
+    fireEvent.changeText(nameInput, 'newUserName')
+    fireEvent.press(submitButton)
+  })
+  expect(nameInput).toHaveProp('value', 'newUserName')
+})
+
+// need to mock FormData? not succeeded so far
```
phillimorland, line 34: As comment says, I tried a number of ways to mock the form data but did not succeed

containers/Reports/__tests__/ReportCreateMessage.test.js
khamui: yes that should be it? the isValid is set by Formik i guess, so you can't change it from "outside"

phillimorland: Yeah I thought that but tried tried a number of other functions then eventually console logged the button props and it's not actually disabled, but now thinking it's because I don't 'fill in' the form, maybe I have to fire an event for that, will give it a last try

goapunk:
@phillimorland I pushed a commit which fixes the async/act warnings and removed the commented tests

containers/Reports/__tests__/ReportCreateMessage.test.js
```diff
@@ -1,71 +1,55 @@
...
 import { testIdeaAdmin, testText } from '../../../tests/TestData'
 import { ReportCreateMessage } from '../ReportCreateMessage'
 
-test('Test ReportCreateMessage Snapshot', () => {
-  const route = { params: {
-    content_type: testIdeaAdmin.content_type,
-    object_pk: testIdeaAdmin.pk
-  }}
-  const { toJSON } = render(<ReportCreateMessage route={route}/>)
+jest.mock('react-native/Libraries/Animated/NativeAnimatedHelper')
```
phillimorland, line 9: is this to ensure the error messages are triggered?

containers/Reports/__tests__/ReportCreateMessage.test.js
```diff
@@ -1,71 +1,55 @@
...
-  fireEvent(submitButton, 'press')
-  const validationErrors = await screen.queryAllByText(
-    /Message must be no longer then 1024 characters/
-  )
-  expect(validationErrors).toBeTruthy()
+  await act(async () => {
+    fireEvent.changeText(messageInput, testText.Lorum1039)
+  })
+  expect(messageInput).toHaveProp('error', 'Message must be no longer then 1024 characters')
+  expect(submitButton).toBeDisabled()
```
phillimorland, line 54: yey!

containers/User/__tests__/SettingsProfile.test.js
```diff
@@ -1,59 +1,52 @@
...
 
 import '@testing-library/jest-native/extend-expect'
 
 import API from '../../../BaseApi'
-import { testUser } from '../../../tests/TestData'
+import { testToken, testUser } from '../../../tests/TestData'
 import { SettingsProfile } from '../SettingsProfile.js'
 
-test('Test SettingsOverview Snapshot no user image', () => {
+test('Test SettingsOverview Snapshot no user image', async () => {
```
phillimorland, line 10: why does this snapshot need async and the others don't?

containers/Reports/__tests__/ReportCreateMessage.test.js
goapunk: no, that's just to remove a warning which was in the terminal

containers/Reports/__tests__/ReportCreateMessage.test.js
goapunk: not great but better than nothing I guess

PR #412, opened 2023-02-20
BaseApi: using optional chaining, returns undefined instead of throwi…

### LINKED ISSUE #411: Remove patch for jest and react-native versions
We have a custom patch to ensure we can run tests with conflicting versions of jest and react-native, however since upgrading (20/02/23) to non conflicting versions, removal of the patch was still not possible. Likely solution is to update babel configs however quick initial attempts failed. 

ref:
https://stackoverflow.com/questions/73550984/invalid-return-value-in-simple-jest-test
https://github.com/facebook/react-native/commit/b5ff26b0b97d6cd600bdb9c33af866971ef14f9c
https://jestjs.io/docs/code-transformation

khamui:
…ng error (esp. in jest-expo).

fixes #411


BaseApi.js
```diff
@@ -4,7 +4,7 @@ import * as Device from 'expo-device'
 export const baseUrl = (() => {
   /*global __DEV__*/
   if (__DEV__) {
-    if (Constants.manifest.extra.localAPI) {
+    if (Constants?.expoConfig?.extra?.localAPI) {
```
goapunk, line 7: the call below in line 27 also needs fixing :)

khamui: oops..

PR #414, opened 2023-02-28
admin: switch from expo buil to eas

goapunk:


philli-m (approved):
works, everything runs but we now need to login to expo in terminal

PR #415, opened 2023-02-28
components/*: replace individual form fields with component and repla…

### LINKED ISSUE #257: create idea long label design looks off
**URL:** https://aplus-dev.liqd.net/teststadt-luca/ideas/2021-00429/update/
**user:** 
**expected behaviour:** labels look the same
**behaviour:** label "maulwurf" breaks the design
**important screensize:**
**device & browser:** Iphone SE, ios 15.2.1
**Comment/Question:** 
This only happens on small screens, on bigger ones it looks fine

IMG_20220203_120600

### LINKED ISSUE #338: style clean up
- see if we can import repeated styles everywhere i.e. container answer: see TextSourceSans
- font sizes: as mobiles use different pixel densities the sizes of our fonts are off almost everywhere compare to what they should be, we need to fix this, possible using https://reactnative.dev/docs/pixelratio.html getFontScale
- once figured out font sizes use variables throughout app, not the multiplyer - already partially implemented

philli-m:
…ce font weight with font to ensure consistency

fixes #257
adding additional components partial fix #338

### DIFF
--- a/components/ButtonCounter.styles.js
+++ b/components/ButtonCounter.styles.js
@@ -1,7 +1,7 @@
 import { StyleSheet } from 'react-native'
 
 import { COLORS } from '../theme/colors'
-import { FONTWEIGHT,LETTERSPACING, LINEHEIGHTS, SIZES } from '../theme/fonts'
+import { FONTS, LETTERSPACING, LINEHEIGHTS, SIZES } from '../theme/fonts'
 
 export const styles = StyleSheet.create({
   container: {
@@ -18,10 +18,10 @@ export const styles = StyleSheet.create({
   },
   highlightUpStyle: {
     color: COLORS.primary,
-    fontWeight: FONTWEIGHT.bold
+    fontFamily: FONTS.familySemiBold
   },
   highlightDownStyle: {
     color: COLORS.danger,
-    fontWeight: FONTWEIGHT.bold
+    fontFamily: FONTS.familySemiBold
   },
 })
--- a/components/ButtonTextInput.js
+++ b/components/ButtonTextInput.js
@@ -4,12 +4,13 @@ import { TouchableOpacity } from 'react-native'
 import IconSLI from 'react-native-vector-icons/SimpleLineIcons'
 
 import { styles } from './ButtonTextInput.styles'
+import { FormLabel } from './FormLabel'
 import { TextSourceSans } from './TextSourceSans'
 
 export const ButtonTextInputFieldContainer = (props) => {
   return (
     <View>
-      <TextSourceSans style={styles.formLabel}>{props.field}</TextSourceSans>
+      <FormLabel>{props.field}</FormLabel>
       {props.children}
     </View>
   )
--- a/components/FormError.js
+++ b/components/FormError.js
@@ -1,0 +1,25 @@
+import React from 'react'
+import { StyleSheet } from 'react-native'
+
+import { COLORS } from '../theme/colors'
+import { SIZES } from '../theme/fonts'
+
+import { TextSourceSans } from './TextSourceSans'
+
+export const FormError = (props) => {
+  const mergedStyles = props.style
+    ? [styles.error, props.style]
+    : styles.error
+  return (
+    <TextSourceSans {...props} style={mergedStyles} accessibilityLiveRegion="polite">
+      {props.children}
+    </TextSourceSans>
+  )
+}
+
+const styles = StyleSheet.create({
+  error: {
+    fontSize: SIZES.xs,
+    color: COLORS.warning
+  }
+})
--- a/components/LabelForm.js
+++ b/components/LabelForm.js
@@ -5,16 +5,17 @@ import { CheckBox } from '@rneui/themed'
 
 import { COLORS } from '../theme/colors'
 
+import { FormError } from './FormError'
+import { FormLabel } from './FormLabel'
 import { styles } from './LabelForm.styles'
-import { TextSourceSans } from './TextSourceSans'
 
 export const LabelListContainer = (props) => {
   return (
     <View style={styles.labelContainer}>
-      <TextSourceSans style={styles.formLabel}>{props.field}</TextSourceSans>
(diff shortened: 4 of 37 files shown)


components/ButtonCounter.styles.js
```diff
@@ -18,10 +18,10 @@ export const styles = StyleSheet.create({
   },
   highlightUpStyle: {
     color: COLORS.primary,
-    fontWeight: FONTWEIGHT.bold
+    fontFamily: FONTS.familySemiBold
   },
   highlightDownStyle: {
     color: COLORS.danger,
-    fontWeight: FONTWEIGHT.bold
+    fontFamily: FONTS.familySemiBold
```
philli-m, line 25: using fontfamily here instead of bold sans text styling as it matches the design

components/ButtonTextInput.js
```diff
@@ -4,12 +4,13 @@ import { TouchableOpacity } from 'react-native'
...
 
 import { styles } from './ButtonTextInput.styles'
+import { FormLabel } from './FormLabel'
 import { TextSourceSans } from './TextSourceSans'
 
 export const ButtonTextInputFieldContainer = (props) => {
   return (
     <View>
-      <TextSourceSans style={styles.formLabel}>{props.field}</TextSourceSans>
+      <FormLabel>{props.field}</FormLabel>
```
philli-m, line 13: using component to ensure consistency

theme/fonts.js
```diff
@@ -1,19 +1,19 @@
-const baseFontSize = 16
-const baseLineHeight = 20
+const baseFontSize = 18
+const baseLineHeight = 22
```
philli-m, line 2: this has helped on the emulators however the problem of how each OS handles font sizes remains

components/LabelForm.js
```diff
@@ -5,13 +5,14 @@ import { CheckBox } from '@rneui/themed'
...
 import { TextSourceSans } from './TextSourceSans'
 
 export const LabelListContainer = (props) => {
   return (
     <View style={styles.labelContainer}>
-      <TextSourceSans style={styles.formLabel}>{props.field}</TextSourceSans>
+      <FormLabel>{props.field}</FormLabel>
       {props.children}
       {props.children.touched &&
         <TextSourceSans style={styles.formError} accessibilityLiveRegion="polite">{props.error}</TextSourceSans>
```
khamui, line 18: would it make sense to have an equivalent component for  form errors? for instance `<FormError>` ?

theme/fonts.js
khamui: can we utilise `Constants.expoConfig` somehow to find out if it is android or iOS and set these values accordingly? would that reduce the difference between both OS?

theme/fonts.js
```diff
@@ -25,8 +25,7 @@ export const LINEHEIGHTS = {
 }
 
 export const FONTWEIGHT = {
-  base: '400',
-  bold: '800',
+  base: '200',
```
khamui, line 28: can we remove it entirely?

components/LabelForm.js
philli-m: @khamui yes! should I add it in this PR do you think? was also thinking containers too?

theme/fonts.js
philli-m: yes, I was originally going to keep using it to reset the checkbox label weight but I ended up removing it in the end

theme/fonts.js
philli-m: That's a good plan although I already read a little on the  but I found when testing with both emulated devices (something I wasn't able to do before) it is a problem for both OS's so thought would up the base anyway

components/LabelForm.js
philli-m: ok added just the form error for now

components/formFields.styles.js
```diff
@@ -1,22 +1,10 @@
...
 
 import { COLORS } from '../theme/colors'
-import { FONTWEIGHT,LINEHEIGHTS, SIZES } from '../theme/fonts'
+import { FONTS, SIZES } from '../theme/fonts'
 import { BORDERWIDTH,SPACINGS } from '../theme/spacings'
 
 export const styles = StyleSheet.create({
-  formTitle: {
-    fontSize: SIZES.md,
-    lineHeight: LINEHEIGHTS.md,
```
philli-m, line 10: wasn't used

containers/Ideas/ExploreListItem.js
```diff
@@ -16,7 +16,10 @@ export const ExploreListItem = (props) => {
   const clockIcon = (<IconSLI name='clock' color={COLORS.grey.medium} />)
 
   return (
-    <TouchableOpacity accessibilityRole="button" onPress={() => props.action(props.item)}>
+    <TouchableOpacity
+      accessibilityRole="button"
+      onPress={() => props.action(props.item)}
+      delayPressIn={100}>
```
philli-m, line 22: I just wanted to see if this would work o proper device, if so one less thing to do but we should maybe look at a refactor using `Pressable`when we look at the story

components/FormError.js
```diff
@@ -0,0 +1,24 @@
...
+import { SIZES } from '../theme/fonts'
+
+import { TextSourceSans } from './TextSourceSans'
+
+export const FormError = (props) => {
+  const mergedStyles = props.style
+    ? [styles.error, props.style]
+    : styles.error
+  return (
+    <TextSourceSans {...props} style={mergedStyles} accessibilityLiveRegion="polite">
```
khamui, line 14: nit-pick: imho the line breaking in `FormLabel` is better

```
<tag>
    <children />
</tag>
```

theme/fonts.js
khamui: ah i see, so it is not only a problem of OS differences...

components/formFields.styles.js
khamui: nice!

khamui (approved):
Good stuff! lgtm...

just a nit-pick :) up to you if you want to do it. Feel free to merge!

PR #416, opened 2023-03-01
WIP refactor image picker (avatar) and issuefix

khamui:


src/containers/User/SettingsProfileAvatar.styles.js
```diff
@@ -0,0 +1,6 @@
+import { StyleSheet } from 'react-native'
+
+export const styles = StyleSheet.create({
+  container: {
```
fuzzylogic2000, line 4: Do you want this as a placeholder? Maybe we can just add it when there are styles?

fuzzylogic2000 (commented):
It does work! Yeah!
I got a warning, though: 
``` WARN  Non-serializable values were found in the navigation state. Check:

SettingsProfileAvatar > params.onGoBack (Function)```
And the tests are failing now. :(

**Review by Rineee — APPROVED** [review]
Yeahh, really nice :partying_face:

**Comment by khamui** [comment]
wip - found another issue (update on ExplorePage)

File: src/containers/Ideas/ExplorePage.js [hunk 2]
```diff
@@ -18,7 +19,15 @@ export const ExplorePage = (props) => {
...
-  const userImage = props?.route?.params?.userImage
+  useEffect(() => {
+    // initially setting profile context
+    setProfileContext({
+      userId: user.pk,
+      userName: user.username,
+      userImage: { uri: user.user_image },
+      userImageFallback: { uri: user.user_image_fallback }
+    })
+  }, [])
```
Rineee, line 31: This might be a stupid question, but I am wondering: we already have the auth context above and then use the user from that context to set the profile context, right? Couldnt we also just use the auth context? Or is that because we dont want to pass all the other stuff?

khamui: Very good question. Actually first I did it in `AuthProvider` but found that pretty packed with stuff. So didn't want to make this bigger. That is why I ended up putting it into its own context, hoping it makes maintaining it easier.

Maybe wrapping the `AuthProvider` with `ProfileProvider` instead of `App`?

Rineee: Haha, I am not sure what that means, but sounds good? :see_no_evil:

Rineee:
@khamui Now I also tested and there is 1 thing that I noticed: when I change the profile picture and then dont click save, but go back to profile page, it updates the picture anyway in context, but not in backend. So when reloading, the old picture is back. Could we change it so it only updates context once save button is clicked?

But apart from that its great and I also think its much better with the context!

src/containers/Ideas/ExplorePage.js
khamui: yes reading the last question again I don't know what I meant. Classic end of day crap...

What I basically tried to say is, I think it is easier to change things later if we have it in its own context, rather than merging it with the AuthContext. Hope that makes more sense...

PR #418, opened 2023-03-02
api: use aplus-app for prod

goapunk:
use new url for prod api


goapunk:
depends on #407

Rineee:
@goapunk What about this one? Only needs rebasing?

goapunk:
@Rineee done

PR #419, opened 2023-03-02
[issues] category dropdown scroll initial attempt

philli-m:


philli-m:
In case anyone gets onto this before I get back: I was hoping we could just add a config but we do so much nesting I think we may have to refactor to simplify our form screens, did not extensively play around with configs though.

src/components/formFields.js
```diff
@@ -68,20 +64,12 @@ export const CheckBoxFormField = (props) => {
...
 
-export const DropdownFormFieldContainer = (props) => {
-  return (
-    <View style={styles.dropdownContainer}>
-      <FormLabel>{props.field}</FormLabel>
-      {props.children}
-    </View>
-  )
-}
-
```
philli-m, line 79: seems to be pointless as dropdown will always need a label

philli-m:
Guy released a beta fix last week :D assume it's ok to use beta till the release anyway right?

khamui:
sweet! nested scroll works for both OSs.

Note: I had to start with clearing cache!

PR #422, opened 2023-03-03
[7152] Jd 2023 03 more translations

goapunk:
depends on #421


khamui:
@goapunk uppsala...needs a rebase and some conflict solving :)

khamui:
not merged yet, due to an error.

Bildschirm­foto 2023-03-09 um 11 22 50

khamui:
@fuzzylogic2000 @Rineee i had an error (see above) when testing this PR. I added a commit which makes it work, but I think @goapunk solution must somehow worked and is actually the way to go. My "fix" has some downsides for instance multi line text is looking off...

fuzzylogic2000:
@khamui YEs, this is very weird! I also had a look and got it working for some of the <Trans> elements, but not for all of them and it does not make any sense. Sometimes even ones that should be straightforward like in Phases does not work. So, I don't know what to do here. :( 
Maybe merge with the removal of the elements and then trying again later?

src/containers/Comments/Comment.js
```diff
@@ -61,16 +63,15 @@ export const Comment = (props) => {
   const getCommentDeleteModalItems = (commentInstance) => {
     return [
       {
-        // space is to center the text
-        title: '   This comment will be deleted.\nThis action cannot be undone.',
+        title: [t('This comment will be deleted.'), t('This action cannot be undone.')],
```
goapunk, line 66: removed the whitespaces as it doesn't work for translated strings anymore (see comment above)

goapunk:
@khamui @fuzzylogic2000 I fixed some issues and made a small change to the modal, this is ready for review again

src/containers/Comments/Comment.js
fuzzylogic2000: Cool! This is much better and on android it's still centered. :)

fuzzylogic2000 (approved):
Cool, works for me (on android).
@khamui I suppose you also want to test iOS?

src/components/ModalItem.js
```diff
@@ -37,9 +37,11 @@ export const ModalItem = (props) => {
       containerStyle={containerStyles}
     >
       <ListItem.Content style={styles.modalContent}>
-        <ListItem.Title style={listItemStyles}>
-          {props.item.title}
-        </ListItem.Title>
+        {titleList.map((t,i) => (
```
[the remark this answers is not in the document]
khamui: nice!

src/containers/Auth/LoginScreen.js
```diff
@@ -74,17 +75,17 @@ export const LoginScreen = () => {
         >
           <View style={styles.container}>
             <TextSourceSans style={styles.title}>
-              Login
+              {t('Login')}
             </TextSourceSans>
-            <Trans>
-              <TextSourceSans style={styles.registerText}>
+            <TextSourceSans style={styles.registerText}>
+              <Trans t={t}>
```
khamui, line 81: is this prop needed?

khamui (approved):
really cool! working fine in iOS!

Just one question to the `<Trans...` tag. Feel free to merge!

src/containers/Auth/LoginScreen.js
goapunk: I  think it's needed to update the view if the language changes, if you just use <trans> it wouldn't. At this point it's probably more of a theoretical problem as you can't really change the language in that view

PR #428, opened 2023-03-14
src: move  getToken to BaseApi, fixes  #370

### LINKED ISSUE #370: move getToken to BaseApi
**URL:** 
**user:** 
**expected behaviour:** 
**behaviour:** 
**important screensize:**
**device & browser:** 
**Comment/Question:** 
We now need the token authentication for every request we are making, so instead of getting the token in the fetch methods, we should move that to BaseApi 

Screenshot?

Rineee:


fuzzylogic2000 (approved):
Supercool!

PR #431, opened 2023-03-16
[7167] Makefile: add publish commands for dev and prod

goapunk:
Add publish commands, the prod channel will use the app a+


fuzzylogic2000 (approved):
NIce! So, both use main to update the channels? Or the branch I am in locally when I use that command?

goapunk:
> NIce! So, both use main to update the channels? Or the branch I am in locally when I use that command?

The branch you are locally in

PR #432, opened 2023-03-16
forms/imageForm/image resize: resizing images if file size is bigger than 5MB and extract into its own hook

### LINKED ISSUE #406: image upload (5MB) not big enough for newer phones
- Result from user testing
- especially for new iphones a problem

khamui:
closes #406


package.json
```diff
@@ -65,7 +65,9 @@
     "react-native-screens": "3.18.0",
     "react-native-web": "0.18.9",
     "sentry-expo": "6.0.0",
-    "yup": "0.32.11"
+    "yup": "0.32.11",
+    "expo-image-manipulator": "11.0.0",
+    "expo-file-system": "15.1.1"
```
khamui, line 70: this is needed, as imagepicker only retrieves file sizes for iOs images, but not for android

fuzzylogic2000 (approved):
Yeah, looks good and works!
Might be a tiny little bit pixelated, though? https://aplus-dev.liqd.net/phillitopictest-org/ideas/2023-00788/

khamui:
> Yeah, looks good and works!
> Might be a tiny little bit pixelated, though? https://aplus-dev.liqd.net/phillitopictest-org/ideas/2023-00788/

it sounds crazy, but this actually does not come from the apps image resizer, but seems to be something happening with the `thumbnail` filter which we use for images on the detail page. I am not sure, but there might be a reason why this was implemented for aplus?  @lthueer do you know?

This idea's image was uploaded from my computer and has a width of 1400x1050 and file size 2.7 MB.

But actually I did not test a lot about the image compression coming from the app. I will test that more and  improve that @fuzzylogic2000

khamui:
@goapunk @philli-m tagging you for review as I pushed a change to this PR

PR #433, opened 2023-03-20
router: switch back to token to check if logged in

goapunk:
I changed it in another PR but realized that doesn't make much sense, so changing it back.

### DIFF
--- a/src/navigation/IdeaNavigator.js
+++ b/src/navigation/IdeaNavigator.js
@@ -33,13 +33,13 @@ const linking = {
 
 export const IdeaNavigator = () => {
   const { t } = useTranslation()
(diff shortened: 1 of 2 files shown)


src/navigation/IdeaNavigator.js
```diff
@@ -33,13 +33,13 @@ const linking = {
 
 export const IdeaNavigator = () => {
   const { t } = useTranslation()
-  const { loading, user } = useAuthorization()
```
khamui, line 36: @goapunk now it throws an error and the profileContext cannot set a user (for username, userId and userImage).

khamui (changes requested):

khamui:
nice!

PR #437, opened 2023-03-20
[#7150] update submit/save btns to be disabled and show loading while processing request

philli-m:


src/components/ButtonSubmit.js
```diff
@@ -13,6 +13,7 @@ export const ButtonSubmit = (props) => {
       accessibilityLabel={props.title}
       accessibilityHint={props.fieldHint}
       disabled={props.disabled}
+      loading={props.loading}
```
philli-m, line 16: I decided to add this so we can see when the button click has been registered clearly to be able to test the performance issue better as have been unable to replicate on the emulators, will test tomorrow in the office

src/containers/Auth/LoginScreen.js
```diff
@@ -27,21 +28,30 @@ export const LoginScreen = () => {
...
+    setSubmitPending(true)
+    return API.postLogin(values)
+      .then((response) => {
+        const {statusCode, data} = response
+        if (statusCode === 200) {
+          signIn(data.token)
+        } else if (statusCode === 400) {
+          let errorMessage
+          if (data.non_field_errors) {
+            setError(data.non_field_errors[0])
```
khamui, line 40: does this work? I think if you handle errors with the catch block below, we should not setError and trigger a rerender but rather `errorMessage = data.non_field_errors[0]`.

maybe it is better to throw the error directly with `throw new Error('some error message string'). With this we immediately enter the `catch` block. Does that make sense?

src/containers/User/SettingsProfile.js
```diff
@@ -103,7 +107,7 @@ export const SettingsProfile = props => {
             <Header transparent={true} navigation={props.navigation} />
             <KeyboardScrollView
               handleSubmit={handleSubmit}
-              isValid={isValid}
+              isValid={isValid && !submitPending}
```
khamui, line 110: nit-pick: maybe we should rename the prop to something like `disabled` or `buttonEnabled`, to be more clear from outside the component?

khamui (commented):
nice! like the pending state. If this happens across more components (comment submission etc.) we may think about a more global state (context) for it?

But for now it is fine, just two comments..

src/containers/Auth/LoginScreen.js
philli-m: did test but i'll check it again, I just wanted it to be same format as our other submits but didn't think too much about it

src/containers/User/SettingsProfile.js
philli-m: after trying this out, it seems that assign disabled to isValid means that the buttons are validated on page load and mean that we don't need to use any additional properties, if using disabled attribute we can use the formik .dirty property in combination with the .isValid property (depending on the form) i feel how we currently do is slightly less confusing

khamui (approved):
Very nice! Oh now there are conflicts..

khamui:
cool!

PR #439, opened 2023-03-21
[#7151] ButtonSubmit//IdeaProject: add submit sticky wrapped component and us…

philli-m:
…e it on IdeaProject

### DIFF
--- a/src/components/ButtonSubmit.styles.js
+++ b/src/components/ButtonSubmit.styles.js
@@ -1,10 +1,11 @@
-import { StyleSheet } from 'react-native'
+import { Dimensions,StyleSheet } from 'react-native'
 
 import { COLORS } from '../theme/colors'
 import { SIZES } from '../theme/fonts'
 import { BORDERRADIUS } from '../theme/spacings'
 
-const submitButtonHeight = 70
+const SCREEN_HEIGHT = Dimensions.get('window').height
+const submitButtonHeight =  SCREEN_HEIGHT * 0.085
 
 export const styles = StyleSheet.create({
   submitButton: {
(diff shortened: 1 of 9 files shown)


src/components/ButtonSubmit.js
```diff
@@ -1,8 +1,17 @@
 import React from 'react'
+import { View } from 'react-native'
 import { Button } from '@rneui/base'
 
 import { styles } from './ButtonSubmit.styles'
 
+export const ButtonSubmitSticky = (props) => {
```
khamui, line 7: personal-pref: Should we put it into its own file? it makes it a tiny little bit easier to spot I think :)

khamui:
cool it works! but there is a greyish area above the button. Is this intended (tested on iPhone 12)?

src/components/ButtonSubmit.js
philli-m: @khamui I kind of thought of it like a container and we use this structure elsewhere like list and form fields, but maybe I should rename to include container?

philli-m: but I also don't really mind, let me know what you prefer and i'll change it :)

philli-m:
> cool it works! but there is a greyish area above the button. Is this intended (tested on iPhone 12)?

no I only tested on a 14, will take a look

src/components/ButtonSubmit.js
khamui: actually I like the `..Sticky` as it makes it very clear what it is. If we never use the `ButtonSubmit` standalone, we can keep it like that and just rename the file to `ButtonSubmitSticky.js`.If we do use it standalone, I would separate it so we have the variant `ButtonSubmit` and `ButtonSubmitSticky`. Does that make sense?

fuzzylogic2000 (approved):
Supercool! 
I don't have a grey shadow and will merge.

PR #440, opened 2023-03-21
[7154] Focus on child-comments

khamui:


src/containers/Comments/Comments.js
```diff
@@ -1,34 +1,63 @@
...
+  const handleOnLayout = (index, commentId, y, height) => {
+    commentViewInfo[index] = {commentId, y, height}
+    setCommentViewInfo(commentViewInfo)
+  }
+
+  // after click on 'reply', this "inbetween" handler will be called to
+  // find the comment which has been clicked on. Then the parent's
+  // handleReply will be fired.
+  const handleReplyLayout = (commentContentType, commentObjectPk) => {
+    const YOFFSET = 80
```
khamui, line 26: offset to bring the parent comment closer to the submit form of child comment

khamui:
wip - on physical devices it does not work properly

src/containers/Comments/Comments.js
```diff
@@ -1,34 +1,62 @@
...
+  // after click on 'reply', this "inbetween" handler will be called to
+  // find the comment which has been clicked on. Then the parent's
+  // handleReply will be fired.
+  const handleReplyLayout = (commentContentType, commentObjectPk) => {
+    const parentComment = commentViewInfo
+      .find(cvc => cvc.commentId === commentObjectPk)
+    props.handleReply(
+      commentContentType,
+      commentObjectPk,
+      parentComment.y + parentComment.height + parentComment.height
```
khamui, line 31: not really sure, why the comment height as to be added twice, but on all tested devices, this works.

src/containers/Ideas/Idea.js
```diff
@@ -366,80 +415,85 @@ export const Idea = (props) => {
...
-              </TextSourceSans>
-            </View>
-            {getLabels().length > 0 && (
-              <View style={styles.labelsContainer}>
-                {getLabels().map((label, idx) => (
-                  <Label key={idx + label} title={label} />
-                ))}
+            <View
+              ref={ideaRef}
+              collapsable={false}
```
khamui, line 420: this "workaround"/"hack" is to prevent an issue with Android.

Described here:
https://github.com/facebook/react-native/issues/29712

philli-m (commented):
just some comments on code, will check on devices now

src/containers/Comments/Comment.js
```diff
@@ -139,9 +139,9 @@ export const Comment = (props) => {
 
   const fetchComment = () => {
     return API.getComment(
-      comment.content_type,
-      comment.object_pk,
-      comment.id,
+      comment?.content_type,
```
goapunk, line 142: dumb question but can they be undefined / null or why do we need ? here ?

src/containers/Comments/Comments.js
philli-m: @khamui sorry dude, this doesn't work when testing with a parent comment with a `read more` link expanded, it focuses on the author of the below comment tested on galaxy s10

khamui: uff yeah, I didn't considered that..will check this case

src/containers/Ideas/Idea.js
```diff
@@ -323,6 +331,52 @@ export const Idea = (props) => {
...
+    // to make sure latest anchorComment is used
+    const onKeyboardDidShow = (e) => {
+      const keyboardHeight = e.endCoordinates.height
+      isReply && scrollViewRef?.current?.scrollTo({
+        x: 0,
+        y: ideaInfoHeight + keyboardHeight + anchorComment
+      })
+      setIsReply(false)
+    }
+    const showSubscription = Keyboard
```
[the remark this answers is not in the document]
goapunk: maybe rather something like `eventHandler` ? then it's `eventHandler.remove()` which seems clearer to me

src/containers/Comments/Comment.js
```diff
@@ -139,9 +139,9 @@ export const Comment = (props) => {
...
+      comment?.object_pk,
+      comment?.id,
```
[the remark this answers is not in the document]
philli-m: sorry that wasn't clear, I meant did we change something that means we have to add ? to check if the comment is there or did it just throw an error before? just curious

src/containers/Comments/Comment.js
khamui: actually good question and also Philli gave me a hint!
It is not needed as we assume if this component is going to be rendered, `comment` can't be undefined/null.

It was a leftover from developing.

src/containers/Ideas/Idea.js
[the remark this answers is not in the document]
khamui: okay as the handler is the callback function `onKeyboard...` I called it `keyboardEventListener`, that should be clear now

src/containers/Ideas/Idea.js
```diff
@@ -323,6 +331,52 @@ export const Idea = (props) => {
...
+
+  useEffect(() => {
+    if (ideaRef.current) {
+      ideaRef.current.measureInWindow((...measures) => {
+        // measures[3] = distance from top to comments list
+        // moving to bottom of screen with minus windowHeight
+        setIdeaInfoHeight(measures[3] - windowHeight)
+      })
+    }
+  }, [ideaInfoHeight])
```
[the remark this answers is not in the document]
khamui: You are right. As we do not trigger this actively `useCallback` does not work for us here. But I changed the dependency to the ref.current.

src/containers/Comments/Comments.js
```diff
@@ -1,34 +1,73 @@
...
+  // after click on 'reply', this "inbetween" handler will be called to
+  // find the comment which has been clicked on. Then the parent's
+  // handleReply will be called.
+  const handleReplyLayout = (commentContentType, commentObjectPk) => {
+    const parentComment = commentViewInfo
+      .find(cvc => cvc.commentId === commentObjectPk)
+    props.handleReply(
+      commentContentType,
+      commentObjectPk,
+      parentComment.y + parentComment.height + getScrollOffset()
```
khamui, line 42: @philli-m i checked the height of the comment form, as you suspected it could be that, but the actual hight of the form is 70 (iOS) and 80 (android). So it would be too less and cutoff weirdly.

With a fixed offset (different for each OS) it works for the comment being expanded (read more / read less).

src/containers/Comments/Comments.js
```diff
@@ -1,34 +1,73 @@
...
 
 import { Comment } from './Comment'
 
+const getScrollOffset = () => {
+  if (Device.osName === 'Android') {
+    return 180
+  } else if (Device.osName === 'iOS') {
+    return 210
+  } else if (Device.osName === 'iPadOS') {
+    return 230
```
khamui, line 16: need to be checked on a real iPad. And eventually adjusted again.

src/containers/Comments/Comments.js
philli-m: something goes a little off here as there is now no way to see the other replies unless you add one yourself

philli-m (commented):
similar issue to before, but only tested on iphone, when a comment is almost off screen, the scroll to doesn't work however my emulator is being buggy in general so will try testing again after a restart but maybe you can also have a look @khamui ? tried on an iphone 12 and 14 and 14 pro

khamui:
> similar issue to before, but only tested on iphone, when a comment is almost off screen, the scroll to doesn't work however my emulator is being buggy in general so will try testing again after a restart but maybe you can also have a look @khamui ? tried on an iphone 12 and 14 and 14 pro
> Checked and the scroll doesn't seem to work on the iphone emulators

@philli-m I could not replicate this issue, but on iPhone simulators you made sure to show the keyboard I guess? Think it is <kbd>command</kbd> + <kbd>k</kbd>.

khamui:
https://github.com/liqd/zt-app/pull/440#discussion_r1155857561

oh this is a bug, it should not focus the child comment anyways, but the parent. Looking into it.

philli-m:
> [#440 (comment)](https://github.com/liqd/zt-app/pull/440#discussion_r1155857561)
> 
> oh this is a bug, it should not focus the child comment anyways, but the parent. Looking into it.

it's not about the focus, it's that the child comments are not shown at all unless the user adds one

philli-m:
> > similar issue to before, but only tested on iphone, when a comment is almost off screen, the scroll to doesn't work however my emulator is being buggy in general so will try testing again after a restart but maybe you can also have a look @khamui ? tried on an iphone 12 and 14 and 14 pro
> > Checked and the scroll doesn't seem to work on the iphone emulators
> 
> @philli-m I could not replicate this issue, but on iPhone simulators you made sure to show the keyboard I guess? Think it is command + k.

right it only works if the keyboard is already opened, will not work if keyboard is closed, even if it has been activated, I assume it'll work on a real device?! can't test so fine by me to be left like this for manual testing

khamui:
https://github.com/liqd/zt-app/pull/440#issuecomment-1494371289

It seems this was never the case. So answering on a comment never showed the "other" child comments. Can you please doublecheck @philli-m ?

philli-m:
@khamui you are entirely right, sorry!

PR #442, opened 2023-03-21
Ks 2023 03 issues

Rineee:


src/BaseApi.js
```diff
@@ -65,6 +65,10 @@ const getHeaders = async (isFormData = false) => {
   if (token) {
     headers['Authorization'] = 'Token ' + token
   }
+  const lang = await AsyncStorage.getItem('language')
+  if (lang && lang == 'de') {
+    headers['Accept-Language'] = lang
+  }
```
Rineee, line 71: We need to send the accept-language header to get the right translation from backend, eg for dates. To avoid to receive eg russian dates, we only send header if language is German.

src/i18n.js
```diff
@@ -25,7 +25,8 @@ export async function loadLanguage() {
         escapeValue: false // react already safes from xss
       },
       keySeparator: '>',
-      nsSeparator: '|'
+      nsSeparator: '|',
+      compatibilityJSON: 'v3',
```
Rineee, line 29: Plurals werent working and I found this issue here: https://github.com/i18next/i18next-scanner/issues/228
As I couldnt make the parser work, I thought it would be fine to switch to back to v3 as we probably will not need complicated plural forms..

goapunk: hmm, weblate also uses v4 for the translations (which probably is fine most of the times). However, as per the issue i18next-scanner seems to be unmaintained / not keeping up with i18next we should probably switch to i18next-parser ? :/

fuzzylogic2000 (approved):
Supercool! Looks good and works.

PR #443, opened 2023-03-22
WIP translations: switch from i18next-scanner to i18next-parser

goapunk:


Rineee:
Cool! Is it ok if I just do that in my PR? Because I also changed the strings a bit there already.. then I would just port if over? @goapunk

locale/en/translation.json
```diff
@@ -1,90 +1,89 @@
...
+    "click to up vote": "click to up vote",
+    "down-votes": "down-votes",
+    "click to down vote": "click to down vote",
+    "Reply": "Reply",
+    "Share": "Share",
+    "Comment must be no longer than 4000 characters.": "Comment must be no longer than 4000 characters.",
+    "Comment input": "Comment input",
+    "Type your comment here": "Type your comment here",
+    "Enter your comment": "Enter your comment",
+    "{{ count }} Entries_one": "{{ count }} Entries",
```
Rineee, line 37: here we need Entry, right?

goapunk: yes

goapunk:
> Cool! Is it ok if I just do that in my PR? Because I also changed the strings a bit there already.. then I would just port if over? @goapunk

yes absolutely, also the way you did the show/hide comments string is better in your PR

locale/de/translation.json
```diff
@@ -8,7 +8,6 @@
     "Username or E-mail address:": "Benutzername oder E-Mail Adresse:",
     "Password:": "Passwort:",
     "Forgot Password?": "Passwort vergessen?",
-    "If you have not created an account yet, then please <1>register</1> first.": "Wenn du noch kein Nutzerkonto angelegt hast, <1>registriere</1> dich bitte zunächst.",
     "Edit": "Bearbeiten",
```
Rineee, line 11: @goapunk Now the Trans component isnt picked up anymore ... ah! https://github.com/liqd/zt-app/blob/d39ca90fa6ce75b06a8f66232d47127ad19c49ae/src/containers/Auth/LoginScreen.js#L81
Do you have an idea how to fix that?

Rineee: Awesome, thanks!

PR #449, opened 2023-03-23
[#7153] Ideas/ExploreListItem//IdeaListItem: add delay to opacity so tiles do…

philli-m:
…n't flicker on scroll


src/containers/Ideas/ExploreListItem.js
```diff
@@ -46,7 +46,7 @@ export const ExploreListItem = (props) => {
     <TouchableOpacity
       accessibilityRole="button"
       onPress={() => props.action(props.item)}
-      delayPressIn={100}>
+      delayPressIn={50}>
```
philli-m, line 49: delay felt a little long when testing

fuzzylogic2000 (approved):
very nice!

PR #461, opened 2023-03-27
fix user settings

fuzzylogic2000:


src/containers/User/SettingsProfile.js
```diff
@@ -29,23 +29,12 @@ export const SettingsProfile = props => {
 
   const makeFormData = (values) => {
     // do not send image data if not changed
-    if (!values['user_image']) {
-      delete values['user_image']
-    } else if (values['user_image'] === profileContext?.newUserImage) {
+    if (!values['user_image'] || !profileContext?.newUserImage
```
khamui, line 32: nice it works! it takes a while to understand all the conditions, though. Maybe remap them to descriptive variables help a bit with readability?

```
    const existingImage = values['user_image']
    const newImage = profileContext?.newUserImage
    const existingIsSameAsNewImage = existingImage === newImage

    if (!existingImage || !newImage || existingIsSameAsNewImage) {
      delete values['user_image']
    } else if (!existingIsSameAsNewImage) {
      // instead of setFieldValue (Formik), assigning "manually" here
      values['user_image'] = profileContext.newUserImage
    }
```

src/containers/User/SettingsProfile.js
```diff
@@ -64,8 +53,23 @@ export const SettingsProfile = props => {
       .then((response) => {
         const {statusCode, data} = response
         if (statusCode == 200) {
-          Alert.alert(t('You\'re profile has been updated'))
+          Alert.alert(t('Your profile has been updated'))
           setSubmitPending(false)
+          // update profile context if username or user_image changed
+          if (values['user_image'] &&
```
khamui, line 59: maybe a little bit more readable?

`if (values?.user_image !== profileContext?.userImage) {..}`

not sure, leave it if its more confusing

khamui (commented):
nice! it works and is much better. Two comments

src/containers/User/SettingsProfile.js
fuzzylogic2000: But that wouldn't work? The first condition is to check if there is user_image in the values and if I got you right `values?.user_image` will tell me if values exist? And I think if it doesn't, it will give `undefined` which will be `!== profileContext?.userImage`, which was also the peoblem why the condition above with the images didn't work and gave a lot of bad requests when we only changed the username and not the image.

khamui: yes you are right! ..sorry!

khamui:
very nice!

PR #464, opened 2023-03-28
containers/Comment: making read more less working for ios by setting …

### LINKED ISSUE #457: [iOS] Read more/ Read less link does not appear in comments
**URL:** 
**user:** 
**expected behaviour:** If a comment text exceeds two lines, there should be a read more/less link to expand
**behaviour:** On iPhones there is no link to expand to see the full comment
**important screensize:**
**device & browser:** iOS devices
**Comment/Question:** 

To figure out a <kbd>Text</kbd> Elements lines, we use the onTextLayout event callback. This function returns different values for Android and for iOS.

Opened an issue in RN:
https://github.com/facebook/react-native/issues/36675

khamui:
…numOfLines dynamically

closes #457


Rineee (changes requested):
Yeah, that works! (Although weird that it doesnt work with the constant, right?)

But also need to do it for the subcomments, there its still broken..

PR #465, opened 2023-03-28
Ks 2023 03 issues issues issues

Rineee:


src/containers/Ideas/Idea.js
```diff
@@ -21,7 +21,6 @@ import { Menu } from '../../components/Menu'
 import { Modal } from '../../components/Modal'
 import { StatusBarStyled } from '../../components/StatusBarStyled'
 import { TextSourceSans } from '../../components/TextSourceSans'
-import { DateService } from '../../services/DateService'
```
khamui, line 24: can we delete this file? or is it still used somewhere?

khamui (commented):
very nice! just one thing

src/containers/Ideas/Idea.js
Rineee: No, it can go, you are right, thanks!

PR #466, opened 2023-03-29
add license, update Makefile

fuzzylogic2000:


goapunk (changes requested):
we usually use AGPL, are we sure we want GPL here?

Rineee (approved):
Cool!

